PR #23, opened 2025-08-13
Function support 

asafpamzn:
Support of functions commands

The tests of PHPRedis are passing.

### DIFF
--- a/valkey_glide_commands_3.c
+++ b/valkey_glide_commands_3.c
@@ -411,8 +411,10 @@ int execute_function_command(zval* object, int argc, zval* return_value, zend_cl
 
             if (result->response) {
                 /* FUNCTION can return various types based on subcommand */
-                status = command_response_to_zval(
-                    result->response, return_value, COMMAND_RESPONSE_NOT_ASSOSIATIVE, false);
+                status = command_response_to_zval(result->response,
+                                                  return_value,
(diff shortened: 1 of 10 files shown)


valkey_glide_commands_3.c
```diff
@@ -602,121 +604,141 @@ int execute_exec_command(zval* object, int argc, zval* return_value, zend_class_
...
+    }
 
-        /* Execute the command */
-        CommandResult* result = execute_command(valkey_glide->glide_client,
-                                                FCall,     /* command type */
-                                                arg_count, /* number of arguments */
-                                                cmd_args,  /* arguments */
-                                                args_len   /* argument lengths */
-        );
+    /* Process args array */
```
jduo, line 677: This block is the same as the one used for keys_array. Do you think it'd be worth putting into a helper?

valkey_glide_commands_3.c
```diff
@@ -602,121 +604,141 @@ int execute_exec_command(zval* object, int argc, zval* return_value, zend_class_
...
-                status = command_response_to_zval(
-                    result->response, return_value, COMMAND_RESPONSE_NOT_ASSOSIATIVE, false);
-                free_command_result(result);
-                return status;
-            }
+    /* Handle the result directly */
+    int status = 0;
+    if (result) {
+        if (result->command_error) {
+            /* Command failed */
```
jduo, line 713: Let's add a TODO to log the error.

PR #36, opened 2025-09-14
Transaction support for all supported commands.

asafpamzn:
Transaction support for all supported commands.


tests/ValkeyGlideBatchTest.php
```diff
@@ -0,0 +1,3854 @@
...
+        $this->assertEquals(-1, $results[0]); // TTL result (no expiration)
+        $this->assertGT(0, $results[1]); // PTTL result (has expiration in ms)
+        $this->assertTrue($results[2]); // PERSIST result (removed expiration)
+
+        // Verify server-side effects
+        $this->assertEquals(-1, $this->valkey_glide->ttl($key1)); // No expiration
+        $this->assertGT(0, $this->valkey_glide->ttl($key2)); // Still has expiration
+        $this->assertEquals(-1, $this->valkey_glide->ttl($key3)); // Expiration removed
+
+        // Cleanup
```
jduo, line 226: We should do cleanup operations in try-finally blocks since they'll get skipped if there are assertion failures.

tests/ValkeyGlideClusterBatchTest.php
```diff
@@ -0,0 +1,192 @@
...
+            TestSuite::errorMessage("Fatal error: %s\n", $ex->getMessage());
+            //TestSuite::errorMessage("Seeds: %s\n", implode(' ', self::$seeds));
+            TestSuite::errorMessage("Seed source: %s\n", self::$seed_source);
+            exit(1);
+        }
+    }
+
+    public function testServerOperationsBatch()
+    {
+
```
jduo, line 51: Is this supposed to be empty?

tests/ValkeyGlideClusterTest.php
```diff
@@ -418,11 +361,10 @@ public function testScan()
 
         /* Iterate over our masters, scanning each one */
         $key_count = $this->valkey_glide->dbsize("allPrimaries");
-
         /* Scan the keys here using ClusterScanCursor - create new cursor each iteration */
         $cursor = new ClusterScanCursor(); // Create fresh cursor each time
         while (true) {
-            $keys = $this->valkey_glide->scan($cursor);
+            $keys = $this->valkey_glide->scan($cursor);
```
jduo, line 367: Trailing whitespace

tests/ValkeyGlideClusterBatchTest.php
asafpamzn: Yes, we can mark it as skipped as the cluster does not support these operations in batch.

tests/ValkeyGlideClusterTest.php
asafpamzn: I will run the linter on PHP.

PR #56, opened 2025-09-27
Initial commit of HFE commands

jduo:
- Add functions and make signatures based on similar functions for non-hash collections
- Add implementation in Zend layer
- Add documentation in stubs
- Add integration tests
- Add batch tests

### Issue link

This Pull Request is linked to issue (URL): [REPLACE ME]

### Checklist

- [ ] This Pull Request is related to one issue.
- [ ] Commit message has a detailed description of what changed and why.
- [ ] Tests are added or updated.
- [ ] CHANGELOG.md and documentation files are updated.
- [ ] Destination branch is correct - main or release
- [ ] Create merge commit if merging release branch into main, squash otherwise.


tests/ValkeyGlideClusterTest.php
```diff
@@ -886,5 +886,523 @@ public function testReplyLiteral()
         $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, false);
     }
 
+    public function testClusterHashExpiration()
```
asafpamzn, line 889: Why do we need different tests for cluster?
Please use the same code for cluster and non cluster. This is what is done for most tests.
Unless there is a cluster specific functionality.

valkey_glide.stub.php
```diff
@@ -1302,6 +1301,402 @@ public function hStrLen(string $key, string $field): ValkeyGlide|int|false;
...
+     *
+     * @param string $key The key of the hash.
+     * @param int $milliseconds The expiration time in milliseconds.
+     * @param string $field The field name.
+     * @param mixed $value The field value.
+     * @param mixed $fields_and_vals Additional field-value pairs.
+     *
+     * @return ValkeyGlide|int|false Returns the number of fields that were added or updated, or false on failure.
+     */
+    public function hPSetEx(string $key, int $milliseconds, string $field, mixed $value, mixed ...$fields_and_vals): ValkeyGlide|int|false;
```
asafpamzn, line 1328: I don't understand why do we need so many functions. I think that it is better to have less functions with more parameters

valkey_glide.stub.php
```diff
@@ -1302,6 +1301,402 @@ public function hStrLen(string $key, string $field): ValkeyGlide|int|false;
...
+     *
+     * @param string $key The key of the hash.
+     * @param int $seconds The expiration time in seconds.
+     * @param string $field The field name.
+     * @param mixed $value The field value.
+     * @param mixed $fields_and_vals Additional field-value pairs.
+     *
+     * @return ValkeyGlide|int|false Returns the number of fields that were added, or false on failure.
+     */
+    public function hSetExNx(string $key, int $seconds, string $field, mixed $value, mixed ...$fields_and_vals): ValkeyGlide|int|false;
```
asafpamzn, line 1367: Again, we can reduce the number of functions

asafpamzn (changes requested):

tests/ValkeyGlideClusterTest.php
```diff
@@ -886,5 +886,523 @@ public function testReplyLiteral()
...
+        $key = $this->createRandomString(10);
+        
+        // Test all HSETEX variants (HSETEX format with field-value pairs)
+        
+        // HPSETEX - milliseconds
+        $result = $this->valkey_glide_cluster->hPSetEx($key, 5000, 'psetex_field', 'psetex_value');
+        $this->assertEquals(1, $result);
+        $this->assertEquals('psetex_value', $this->valkey_glide_cluster->hGet($key, 'psetex_field'));
+        $pttl = $this->valkey_glide_cluster->hPTtl($key, 'psetex_field');
+        $this->assertGreaterThan(4000, $pttl[0]);
```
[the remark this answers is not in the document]
jduo: I've reworked these tests such that they take in the timestamp of the expiry-setting call to calculate the elapsed time automatically. The caller specifies the range for the buffer. This would make it so we don't need to define variables but still handle timing differences due to different test speeds.

valkey_glide_hash_common.c
```diff
@@ -538,6 +584,372 @@ int prepare_h_randfield_args(h_command_args_t* args,
...
+            char* expiry_str                           = safe_format_int(args->expiry);
+            (*allocated_strings)[(*allocated_count)++] = expiry_str;
+            (*args_out)[arg_idx]                       = (uintptr_t) expiry_str;
+            (*args_len_out)[arg_idx]                   = strlen(expiry_str);
+            arg_idx++;
+        }
+    }
+
+    /* Add "FIELDS" keyword */
+    const char* fields_keyword = "FIELDS";
```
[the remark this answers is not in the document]
jduo: Duplication should be reduced alot.

PR #174, opened 2026-03-26
PHP: fix memory leaks and resource management issues

prateek-kumar-improving:
### Summary

Fixes multiple memory leaks, a potential crash in debug builds, and a resource management gap.

### Issue link

This Pull Request is linked to issue (URL): https://github.com/valkey-io/valkey-glide-php/issues/175

- ValkeyGlide::close() and ValkeyGlideCluster::close() now actually release the underlying Rust client connection instead of being no-ops. Previously, connections were only released on garbage collection.

### Implementation

| # | Fix | File(s) |
|---|-----|---------|
| 1 | Check EG(exception) after _initialize_open_telemetry() | valkey_glide.c |
| 2 | Extract cleanup_route() helper and replace bare return NULL with cleanup_route() | command_response.c |
| 3 | Extract validate_command_params() helper to consolidate parameter validation in execute_command_with_route() | command_response.c |
| 4 | NULL out route.key and reset key_allocated after free in cleanup_route() to prevent potential double-free | command_response.c |
| 5 | Extract valkey_glide_clear_batch_state() as a shared non-static helper | common.h, valkey_glide_commands_3.c |
| 6 | Call valkey_glide_clear_batch_state() in free_valkey_glide_object() | valkey_glide.c |
| 7 | Call valkey_glide_clear_batch_state() in close() | valkey_glide.c, valkey_glide_cluster.c |
| 8 | Implement close() to call close_glide_client()  | valkey_glide.c, valkey_glide_cluster.c |

### Testing
All integration tests run.

### Checklist

### DIFF
--- a/command_response.c
+++ b/command_response.c
@@ -237,6 +237,42 @@ uint8_t* create_route_bytes_from_route(cluster_route_t* route, size_t* route_byt
 }
 
 /* Execute a command and handle common error checking */
+/* Helper to free route-related allocations */
+static void cleanup_route(uint8_t* route_bytes, cluster_route_t* route) {
+    if (route_bytes) {
+        efree(route_bytes);
+    }
+    if (route->type == ROUTE_TYPE_KEY && route->data.key_route.key_allocated) {
+        efree(route->data.key_route.key);
+    }
+}
+
+/* Validate parameters before FFI call, returns 1 if valid, 0 otherwise */
+static int validate_command_params(const void*          glide_client,
+                                   unsigned long        arg_count,
+                                   const uintptr_t*     args,
+                                   const unsigned long* args_len) {
+    if (!glide_client) {
+        VALKEY_LOG_ERROR("parameter_validation", "glide_client is NULL");
+        return 0;
+    }
+
+    if (arg_count > 0) {
+        if (!args) {
+            VALKEY_LOG_ERROR_FMT(
+                "parameter_validation", "args is NULL but arg_count is %lu", arg_count);
+            return 0;
+        }
+        if (!args_len) {
+            VALKEY_LOG_ERROR_FMT(
+                "parameter_validation", "args_len is NULL but arg_count is %lu", arg_count);
+            return 0;
+        }
+    }
+
+    return 1;
+}
+
 CommandResult* execute_command_with_route(const void*          glide_client,
                                           enum RequestType     command_type,
                                           unsigned long        arg_count,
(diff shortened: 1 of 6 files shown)


valkey_glide.c
```diff
@@ -568,6 +572,27 @@ ZEND_GET_MODULE(valkey_glide)
 void free_valkey_glide_object(zend_object* object) {
     valkey_glide_object* valkey_glide = VALKEY_GLIDE_PHP_GET_OBJECT(valkey_glide_object, object);
 
+    /* Free buffered batch commands if object is destroyed mid-batch */
```
jeremyprime, line 585: Should the buffered commands also be cleaned up when `close` is called?  This could be moved to a helper function and called here and in `close`.

This also applies to `valkey_glide_cluster.c` file.

valkey_glide_str_commands.c
```diff
@@ -454,8 +454,10 @@ int execute_sort_command(zval* object, int argc, zval* return_value, zend_class_
         /* Free the argument arrays */
         efree(args);
         efree(args_len);
-        efree(offset_str);
-        efree(count_str);
+        if (offset_str)
```
jduo: Isn't efree(NULL) a no-op? Why add these if statements for just these two?

valkey_glide.c
```diff
@@ -568,6 +572,27 @@ ZEND_GET_MODULE(valkey_glide)
...
+    if (valkey_glide->buffered_commands) {
+        size_t i, j;
+        for (i = 0; i < valkey_glide->command_count; i++) {
+            struct batch_command* cmd = &valkey_glide->buffered_commands[i];
```
jduo: Do we need to free cmd as well? Or is buffered_commands allocated as one big array of batch_commands?

command_response.c
```diff
@@ -237,6 +237,42 @@ uint8_t* create_route_bytes_from_route(cluster_route_t* route, size_t* route_byt
 }
 
 /* Execute a command and handle common error checking */
+/* Helper to free route-related allocations */
+static void cleanup_route(uint8_t* route_bytes, cluster_route_t* route) {
+    if (route_bytes) {
+        efree(route_bytes);
+    }
+    if (route->type == ROUTE_TYPE_KEY && route->data.key_route.key_allocated) {
+        efree(route->data.key_route.key);
```
jduo, line 246: Should this set route->data.key_route.key to NULL so it doesn't get double-freed? Do we need to clean up anything else on route, or clean up route itself?

jduo (changes requested):

valkey_glide.c
prateek-kumar-improving: Updated. Created a helper method: valkey_glide_clear_batch_state

valkey_glide_str_commands.c
prateek-kumar-improving: Updated.

command_response.c
prateek-kumar-improving: Updated.
Set `route->data.key_route.key = NULL;` and updated `route->data.key_route.key_allocated = 0;`
The only dynamically allocated member is key_route.key so, no other cleanup required.

valkey_glide.c
prateek-kumar-improving: We don't need to free cmd as well. 
buffered_commands is allocated as one contiguous array via ecalloc():

```
// valkey_glide_commands_3.c:845
valkey_glide->buffered_commands = (struct batch_command*) ecalloc(
    initial_capacity, sizeof(struct batch_command));
```
Each cmd is just a pointer into that array (&valkey_glide->buffered_commands[i]), not a separate allocation.

This code is now used from an already existing method now renamed to valkey_glide_clear_batch_state()

jduo (approved):

jeremyprime (approved):

PR #196, opened 2026-05-12
feat: add address resolver support for standalone and cluster clients

affonsov:
### Summary

Adds support for a custom address resolver callback in both `ValkeyGlide` (standalone) and `ValkeyGlideCluster` clients. The resolver allows users to intercept host/port resolution at connection time — useful for DNS overrides, service mesh scenarios, or testing with fake addresses.

### Issue link

This Pull Request is linked to issue (URL): [REPLACE ME]

- `ValkeyGlide::connect()` accepts a new optional `?callable $address_resolver = null` parameter.
- `ValkeyGlideCluster::__construct()` accepts a new optional `?callable $address_resolver = null` parameter.
- The callable receives `(string $host, int $port)` and should return a `['host' => ..., 'port' => ...]` array. If it throws or returns invalid data, the client falls back to the original address.
- Resolver state is cleared on module shutdown (`MSHUTDOWN`) and request shutdown (`RSHUTDOWN`).

### Implementation

- New `valkey_glide_address_resolver.c/.h`: implements `valkey_glide_set_address_resolver`, `valkey_glide_clear_address_resolver`, and the C callback `valkey_glide_address_resolver_callback` that bridges the PHP callable into the Rust FFI `create_client` call.
- `common.h`: added `zval* address_resolver` field to `valkey_glide_base_client_configuration_t` and `valkey_glide_php_common_constructor_params_t`.
- `config.m4`: registered `valkey_glide_address_resolver.c` in the extension source list.
- `valkey_glide.c`: wires the resolver through `valkey_glide_create_connection`; adds `PHP_RSHUTDOWN` to clear resolver state per request.
- `valkey_glide_core_commands.c`: passes `address_resolver_cb` to `create_client` when a resolver is set.
- `valkey_glide_cluster.c`, `src/client_constructor_mock.c/.stub.php`: same parameter threading for cluster and mock constructors.

### Limitations

None known. The test suite skips address resolver tests when running in TLS mode due to test setup constraints (the tests use plain connections without TLS config), not a feature limitation.

### Testing

Added `tests/AddressResolverTest.php` with four integration tests:
- `testAddressResolverWithFakeAddress` — resolver maps a fake host to the real server; asserts successful connection.
- `testAddressResolverExceptionFallsBackToOriginal` — resolver throws `RuntimeException`; verifies fallback to original address.
- `testAddressResolverReturnsInvalidFallsBackToOriginal` — resolver returns `null`; verifies fallback.
- `testNullAddressResolverConnectsNormally` — null resolver; normal connection path.

### Checklist

- [ ] This Pull Request is related to one issue.
- [ ] Create merge commit if merging release branch into main, squash otherwise.


tests/AddressResolverTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Resolver throws an exception — client falls back to original address, ping() succeeds.
+     */
+    public function testAddressResolverExceptionFallsBackToOriginal()
+    {
+        $this->skipIfTlsEnabled();
+
+        $resolver = function (string $host, int $port): array {
+            throw new RuntimeException("resolver error");
```
jduo, line 47: Can we change this test to better prove the resolver ran? Like having the lambda take in a local variable that gets mutated before throwing the exception.

valkey_glide_address_resolver.c
```diff
@@ -0,0 +1,85 @@
...
+#include "valkey_glide_address_resolver.h"
+
+#include <string.h>
+#include <zend_exceptions.h>
+
+static zval g_address_resolver;
+static bool g_has_address_resolver = false;
+
+void valkey_glide_set_address_resolver(zval* callable) {
+    if (g_has_address_resolver) {
```
jduo: In the Rust code, is the address resolver also global?

tests/AddressResolverTest.php
```diff
@@ -0,0 +1,95 @@
...
+        );
+
+        $this->assertConnected($client);
+        $client->close();
+    }
+
+    /**
+     * Resolver throws an exception — client falls back to original address, ping() succeeds.
+     */
+    public function testAddressResolverExceptionFallsBackToOriginal()
```
prateek-kumar-improving, line 40: Do we need a similar test for cluster client?

valkey_glide_address_resolver.c
```diff
@@ -0,0 +1,85 @@
+/** Copyright Valkey GLIDE Project Contributors - SPDX-Identifier: Apache-2.0 */
+
+#include "valkey_glide_address_resolver.h"
+
+#include <string.h>
+#include <zend_exceptions.h>
+
+static zval g_address_resolver;
```
prateek-kumar-improving: What will happen in the case if multiple clients are created in a single process and then one of clients reconnects? This will always use the resolver assigned using the last created client.

Why not create a per client address resolver? Storing it in valkey_glide_object struct should work?

tests/AddressResolverTest.php
affonsov: Updated the test

tests/AddressResolverTest.php
affonsov: Added cluster tests

valkey_glide_address_resolver.c
affonsov: no, I refactored the code
replaced the process-global with per-client libffi closures (same mechanism Python-sync uses via ffi.callback()). Each client gets a unique C function pointer via ffi_closure_alloc + ffi_prep_closure_loc.

valkey_glide_address_resolver.c
affonsov: Refactored the code 
replaced the process-global with per-client libffi closures (same mechanism Python-sync uses via ffi.callback()). Each client gets a unique C function pointer via ffi_closure_alloc + ffi_prep_closure_loc.

prateek-kumar-improving (approved):

jduo (approved):

PR #209, opened 2026-05-22
PHP: add PHPRedis compatibility check script

prateek-kumar-improving:
### Summary

Adds a script (compatibility_check.php) that generates a compatibility report comparing the Valkey GLIDE PHP client against PHPRedis for both standalone (Redis vs ValkeyGlide) and 
cluster (RedisCluster vs ValkeyGlideCluster) APIs.

### Issue link

This Pull Request is linked to issue (URL): No issue

The script categorizes every PHPRedis command into four groups:
1. Compatible APIs — Implemented in Glide with matching PHPRedis signature
2. Not Compatible APIs — Present in both but with different argument signatures
3. Not Supported in Valkey — PHPRedis commands for features not in the Valkey engine (Redis 8+, KeyDB, vector sets)
4. Not Implemented — Commands supported by Valkey and present in PHPRedis, but not yet implemented in Glide

Outputs a summary to stdout and writes a detailed COMPATIBILITY_REPORT.md.

### Current results

| | Standalone | Cluster |
|---|---|---|
| Compatible | 75.6% | 83.3% |
| Not Compatible | 4.3% | 4.8% |
| Not Supported in Valkey | 9.3% | 3.5% |
| Not Implemented | 10.9% | 8.3% |

### How to run

bash
php compatibility_check.php

### Checklist

- [x] This Pull Request is related to one issue - No issue


compatibility_check.php
```diff
@@ -0,0 +1,554 @@
...
+];
+
+// ============================================================================
+// Commands with known incompatibilities (different signature/behavior)
+// These exist in both but have meaningful differences
+// ============================================================================
+
+$not_compatible = [
+    // PHPRedis hexpire/hpexpire/hexpireat/hpexpireat take (key, ttl, fields_array, mode)
+    // Glide takes (key, ttl, mode, field, ...other_fields) - different argument order
```
jduo: Is it just order that is incorrect here or is there another reason? We should just rework the implementations to be consistent. (Maybe we make the C implementation take in the PHP-redis style ordering).

compatibility_check.php
```diff
@@ -0,0 +1,554 @@
...
+    // Glide takes (key, fields_array, options) - similar but options format differs
+    'hgetex',
+
+    // PHPRedis hsetex takes (key, fields_array, expiry)
+    // Glide has hSetEx with different signature (key, seconds, mode, field, value, ...)
+    'hsetex',
+
+    // PHPRedis hgetdel takes (key, fields_array)
+    // Glide doesn't have hgetdel (it has hGetDel but not in the stub - actually not listed)
+    // Actually let's check - removing from here if not implemented
```
jduo: What does this mean?

compatibility_check.php
prateek-kumar-improving: yes, it is just the order and variadic field types. It is possible to fix this with refactoring.

Also, added the exact differences in the signatures in the comments for each commands.

compatibility_check.php
prateek-kumar-improving: Was checking the case for this command separately. It got added as a comment instead of getting removed from the list. Cleaned up the comment by removing it in the next commit.

compatibility_check.php
jduo: We need to maintain compatibility both with what PHPRedis's signatures are, and with what we're currently supporting (since users may already be using it). Let's create Issues to fix the PHPRedis compatibility.

prateek-kumar-improving: Created an issue: https://github.com/valkey-io/valkey-glide-php/issues/215

jduo (approved):

compatibility_check.php
```diff
@@ -0,0 +1,550 @@
...
+
+    // Redis 8+ specific commands not in Valkey
+    'msetex',       // MSETEX - Redis 8+
+    'delex',        // DELEX - Redis 8+
+    'delifeq',      // DELIFEQ - Valkey 9+ (but not standard PHPRedis target)
+    'xdelex',       // XDELEX - Redis 8+
+    'digest',       // DIGEST - Redis 8+
+    'getWithMeta',  // GETWITHMETA - Redis 8+
+    'hGetWithMeta', // HGETWITHMETA - Redis 8+
+    'waitaof',      // WAITAOF - Redis 7.2+ (available in Valkey but not core PHPRedis compat target)
```
jamesx-improving: maybe tweak the wording here a bit: the array name, as well as the comment of the entire array on top says "not supported in valkey", but we actually have commands that IS supported in valkey but not in PHPRedis, judging by the individual comments of the `delifeq` and `waitaof` commands.

compatibility_check.php
```diff
@@ -0,0 +1,550 @@
...
+
+    // Misc
+    'sAddArray', 'pconnect',
+];
+
+// ============================================================================
+// Glide PHP implemented methods (from valkey_glide.stub.php)
+// ============================================================================
+
+$glide_commands = [
```
jamesx-improving: Hardcoded command list might get outdated as development progresses. Maybe for a future PR but please consider derive them dynamically using reflection.

jamesx-improving (approved):
LGTM. Two non-blocking comments for the author's consideration.

compatibility_check.php
prateek-kumar-improving: Updated.

compatibility_check.php
prateek-kumar-improving: Updated.

Issue #218, opened 2026-06-03
Symfony Lock RedisStore fatals on every acquire(): getLastError()/clearLastError() not implemented


tnucera:
### Summary

With the PHPRedis aliases enabled, **Symfony Lock breaks on every lock acquisition**: `ValkeyGlide` does not implement `clearLastError()` / `getLastError()`, which Symfony's `Lock\Store\RedisStore::evaluate()` calls unconditionally. Reproduced live below.

### Environment (reproduced)

- valkey-glide-php **1.1.0** (GitHub release asset; not on PECL — `pecl.php.net/get/valkey_glide-1.1.0.tgz` → 404)
- PHP **8.3.31** (`php:8.3-cli-bookworm`)
- symfony/lock **v7.4.9**
- Valkey **8** (single node)

### Reproduction

```php
\ValkeyGlide::registerPHPRedisAliases();                       // => true
$redis = new \Redis();                                         // aliased to ValkeyGlide
$redis->connect(addresses: [['host' => 'valkey', 'port' => 6379]]);

$store = new \Symfony\Component\Lock\Store\RedisStore($redis);
$lock  = (new \Symfony\Component\Lock\LockFactory($store))->createLock('demo');
$lock->acquire();
```

Output:

```
ValkeyGlide::getLastError exists:   false
ValkeyGlide::clearLastError exists: false
registerPHPRedisAliases(): true
connected, instanceof \Redis: true
--- calling lock->acquire() ---
THROWN: Error: Call to undefined method ValkeyGlide::clearLastError()
at vendor/symfony/lock/Store/RedisStore.php:236
```

### Root cause

`RedisStore::evaluate()` (symfony/lock 7.4, lines 235–236) enters its phpredis branch for any object that is `instanceof \Redis` — which an aliased `ValkeyGlide` satisfies — and immediately calls `clearLastError()`:

```php
if ($this->redis instanceof \Redis || ... || $this->redis instanceof \RedisCluster) {
    $this->redis->clearLastError();                 // line 236 — undefined on ValkeyGlide
    $result = $this->redis->evalSha($scriptSha, array_merge([$resource], $args), 1);
    if (null !== ($err = $this->redis->getLastError()) && str_starts_with($err, self::NO_SCRIPT_ERROR_MESSAGE_PREFIX)) {
        ...                                          // getLastError() also undefined
    }
}
```

[...]

PR #221, opened 2026-06-03
PHP: Implement getLastError() and clearLastError() for PHPRedis compatibility

tnucera:
### Summary

`ValkeyGlide` / `ValkeyGlideCluster` did not implement the PHPRedis error-introspection methods `getLastError()` / `clearLastError()`. PHPRedis-compatible libraries depend on them: Symfony Lock `RedisStore::evaluate()` calls `clearLastError()` unconditionally on **every** lock acquire, which fataled with `Call to undefined method ValkeyGlide::clearLastError()` (#218, reproduced live).

The extension's error model already returns `false` to the caller on command errors — but the server error message produced by glide-core was discarded at the execution error sites. This PR stores that already-produced message on the client object and exposes the two accessors. **No behavior change on success paths or on paths that throw `ValkeyGlideException`.**

### Issue link

This Pull Request is linked to issue (URL): https://github.com/valkey-io/valkey-glide-php/issues/218

- New methods on both `ValkeyGlide` and `ValkeyGlideCluster`, with PHPRedis semantics:
  - `getLastError(): ?string` — last recorded command error, `NULL` if none
  - `clearLastError(): bool` — resets it, always returns `true`
  - errors overwrite the previous one; successful commands do **not** clear it; only `clearLastError()` resets
- Command errors that are converted into a `false` return now record the server error message on the object. Paths that throw `ValkeyGlideException` are unchanged (the exception already carries the message).

### Implementation

- `zend_string* last_error` field on `valkey_glide_object` (shared by both classes), released in `free_valkey_glide_object`
- One shared helper, `valkey_glide_record_command_error()`, called at every execution path that converts a command error into `false`: the core command framework (`execute_core_command`), the script/eval validation macros, batch `exec()`, and the z / geo / list / hash / stream / set / function family executors (these have their own execution paths that bypass the core framework and previously dropped the error silently)
- Methods implemented via `GETLASTERROR_METHOD_IMPL` / `CLEARLASTERROR_METHOD_IMPL` macros following the existing `GETOPTION_METHOD_IMPL` pattern; stubs updated for both classes (arginfo is regenerated at build time)

**Reviewer attention — NOSCRIPT normalization** (`valkey_glide_set_last_error`, valkey_glide.c): glide-core (redis-rs) reformats server errors as `An error was signalled by the server: - NoScriptError: No matching script.`, while PHPRedis consumers pattern-match the raw RESP code — Symfony Lock does `str_starts_with($err, 'NOSCRIPT')` to reload missing scripts, which is the standard evalSha-first flow. Without restoring the prefix, the fallback never triggers and every `RedisStore` acquire fails. The helper strips the glide prefix and restores the `NOSCRIPT` code for that error kind. This is string-based because the FFI `RequestErrorType` enum only exposes Unspecified/ExecAbort/Timeout/Disconnect — exposing the full error kind through the FFI would be the robust long-term fix (happy to open a follow-up issue on valkey-glide core); a regression test guards the current mapping.

### Limitations

- Only the `NOSCRIPT` RESP code is restored; other server errors keep glide-core's message format (e.g. `WRONGTYPE: ...` instead of the raw `WRONGTYPE ...`). `str_contains`/prefix checks still match for the common codes.
- `last_error` is not populated for connection/lifecycle errors that throw — by design, the exception carries the message.

### Testing

- New `testGetLastErrorAndClearLastError` in `tests/ValkeyGlideTest.php` (inherited by `ValkeyGlideClusterTest`; uses `{hashtag}` keys so it is cluster-safe): fresh-client NULL, error recording via eval and via the core path, persistence across successful commands, clear semantics, overwrite semantics, NOSCRIPT prefix regression, and a per-family loop (zAdd / rPush / hSet / xAdd / geoadd / sAdd against a wrong-type key)
- Full local suite (extension built from this branch): 206 passed / 3 failed / 12 skipped — identical profile to `main`: the 3 failures are the `localhost`-hardcoded connection tests (pass when the test server shares the network namespace) and the 12 skips are the Valkey 9 hash-field-expiration gates (all pass against `valkey:9`)
- End-to-end with Symfony Lock 7.4 + PHP 8.3 + Valkey 8 (the exact #218 reproduction): `acquire()` → `true`, `refresh()`, `release()`, contention (second acquire `false`, no fatal), and the NOSCRIPT script-reload path all work; 22/22 checks pass
- `clang-format-18` and `phpcs` clean

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * PHP: Fix pie install from Packagist using PIE pre-packaged-source download method ([#233](https://github.com/valkey-io/valkey-glide-php/pull/233))
 * PHP: Fix pie install from Packagist by setting preferred-install to source for submodule resolution ([#232](https://github.com/valkey-io/valkey-glide-php/pull/232))
 * PHP: Fix pie install ([#229](https://github.com/valkey-io/valkey-glide-php/pull/229))
+* PHP: Implement getLastError()/clearLastError() for PHPRedis compatibility ([#221](https://github.com/valkey-io/valkey-glide-php/pull/221))
 * PHP: Set CLIENT SETINFO lib-name=GlidePHP and lib-ver to extension version for proper client identification
 * PHP: Add address resolver support for standalone and cluster clients, allowing custom host/port remapping at connection time ([#196](https://github.com/valkey-io/valkey-glide-php/pull/196))
 
--- a/valkey_glide_cluster.stub.php
+++ b/valkey_glide_cluster.stub.php
@@ -299,6 +299,16 @@ class ValkeyGlideCluster
     public function getOption(int $option): mixed;
 
     /**
+     * @see ValkeyGlide::getLastError()
+     */
+    public function getLastError(): ?string;
+
+    /**
+     * @see ValkeyGlide::clearLastError()
+     */
+    public function clearLastError(): bool;
+
+    /**
      * @see ValkeyGlide::append()
      */
     public function append(string $key, mixed $value): ValkeyGlideCluster|bool|int;
--- a/valkey_glide_s_common.c
+++ b/valkey_glide_s_common.c
@@ -840,7 +840,8 @@ int execute_s_generic_command(valkey_glide_object* valkey_glide,
         /* Always run the processor: on the SCAN path it resets the scan iterator
          * and frees the emalloc'd scan_data, which would otherwise leak on error.
          * The processors handle a NULL response by returning a false/empty value. */
-        status = process_result(result->command_error ? NULL : result->response, scan_data, return_value);
+        status = process_result(
+            result->command_error ? NULL : result->response, scan_data, return_value);
     }
     free_command_result(result);
 
--- a/valkey_glide_s_common.c
+++ b/valkey_glide_s_common.c
@@ -834,7 +834,11 @@ int execute_s_generic_command(valkey_glide_object* valkey_glide,
     /* Execute the command synchronously */
     result = execute_command(valkey_glide->glide_client, cmd_type, arg_count, cmd_args, args_len);
     if (result) {
-        status = process_result(result->response, scan_data, return_value);
+        if (result->command_error) {
+            valkey_glide_record_command_error(valkey_glide, result);
+        } else {
+            status = process_result(result->response, scan_data, return_value);
+        }
     }
     free_command_result(result);
 
--- a/valkey_glide_s_common.c
+++ b/valkey_glide_s_common.c
@@ -836,9 +836,11 @@ int execute_s_generic_command(valkey_glide_object* valkey_glide,
     if (result) {
         if (result->command_error) {
             valkey_glide_record_command_error(valkey_glide, result);
-        } else {
-            status = process_result(result->response, scan_data, return_value);
         }
+        /* Always run the processor: on the SCAN path it resets the scan iterator
+         * and frees the emalloc'd scan_data, which would otherwise leak on error.
+         * The processors handle a NULL response by returning a false/empty value. */
+        status = process_result(result->command_error ? NULL : result->response, scan_data, return_value);
     }
     free_command_result(result);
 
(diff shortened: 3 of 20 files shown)


valkey_glide_s_common.c
```diff
@@ -834,7 +834,11 @@ int execute_s_generic_command(valkey_glide_object* valkey_glide,
     /* Execute the command synchronously */
     result = execute_command(valkey_glide->glide_client, cmd_type, arg_count, cmd_args, args_len);
     if (result) {
-        status = process_result(result->response, scan_data, return_value);
+        if (result->command_error) {
+            valkey_glide_record_command_error(valkey_glide, result);
+        } else {
+            status = process_result(result->response, scan_data, return_value);
```
prateek-kumar-improving: Why should this method not be called in case the value of result->command_error is true?

valkey_glide_cluster.stub.php
```diff
@@ -297,6 +297,16 @@ public function setOption(int $option, mixed $value): bool;
      */
     public function getOption(int $option): mixed;
 
+    /**
+     * @see ValkeyGlide::getLastError()
+     */
+    public function getLastError(): ?string;
```
prateek-kumar-improving, line 304: Please confirm if the method names are camel case in PHPRedis. If not please add additional signatures with common implementation.

valkey_glide_s_common.c
tnucera: Good catch — fixed in e1a1b3b. The original code called `process_result` unconditionally; moving it into the `else` branch introduced a regression on the SCAN path: `scan_data` is `emalloc`'d and freed only inside `process_s_scan_result_async` (even for a NULL response), and the `cleanup:` label doesn't free it — so skipping the processor leaked `scan_data` and left the scan cursor unreset. The processor now always runs (with a NULL response on error, which it handles by returning a false/empty value), and the command error is still recorded.

valkey_glide_cluster.stub.php
tnucera: Confirmed: PHPRedis declares these as `getLastError()` and `clearLastError()` (camelCase only — there are no snake_case aliases), so the signatures match the PHPRedis API as written. This is also consistent with the canonical method list in `compatibility_check.php` (#209). No additional signatures needed.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+* PHP: Implement getLastError()/clearLastError() for PHPRedis compatibility — command errors now record the server error message on the client object (fixes Symfony Lock RedisStore fatal) ([#221](https://github.com/valkey-io/valkey-glide-php/pull/221))
```
affonsov: nit: changelog entry can be simplified, don't need explanations here

affonsov (commented):
LGTM
Please fix lint errors

tnucera:
Rebased onto latest `main` (only conflict was CHANGELOG ordering). Both review points addressed:

- **SCAN path** (`valkey_glide_s_common.c`): the processor now runs unconditionally again (NULL response on error), so `scan_data` is freed and the scan cursor reset even on failure, while the command error is still recorded. Verified: `sScan` on a wrong-type key returns `false`, `getLastError()` reports `WRONGTYPE`, and 100 consecutive failing scans run with no crash or leak.
- **Naming**: `getLastError`/`clearLastError` are camelCase-only in PHPRedis (no snake_case aliases), matching the canonical list in `compatibility_check.php` (#209).

Ready for review — and merge whenever it looks good to you.

tnucera:
Done — simplified the CHANGELOG entry to just the one-liner.

prateek-kumar-improving (approved):

prateek-kumar-improving:
Please fix lint errors. 
Check logs here for more details about the lint errors: https://github.com/valkey-io/valkey-glide-php/actions/runs/28925879357/job/86734021966?pr=221

affonsov (approved):

prateek-kumar-improving:
Fixed the CI lint error in this commit: https://github.com/valkey-io/valkey-glide-php/pull/221/changes/ea2ffa9d0538e408db7935fede34aee1f50c5173

Issue #224, opened 2026-06-11
Fix PIE install for PHP Glide Client


prateek-kumar-improving:
1. Fix pie install command.
2. If required, use this to test: https://github.com/opensource-for-valkey/valkey-glide-php-docker

prateek-kumar-improving:
For making our build work, encountered other existing issues with pie and had to fix those issues as well.
Other open source projects have also faced related issues: https://github.com/php/pie/issues/39.
Had to make an update in existing PHP release workflows to account for some of these pie issues.

Existing issues fixed here: https://github.com/valkey-io/valkey-glide-php/issues/240
pie install fixed in this release: https://github.com/valkey-io/valkey-glide-php/releases#release-v1.1.2

PR #229, opened 2026-06-22
PHP: Fix pie install

prateek-kumar-improving:
### Summary

Fixes the build system to correctly resolve the valkey-glide submodule commit during pie install, instead of falling back to main HEAD.

Added .github/workflows/test-pie.yml to validate pie install works.

### Issue link

This Pull Request is linked to issue (URL): https://github.com/valkey-io/valkey-glide-php/issues/224

### Problem

When pie install valkey-io/valkey-glide-php runs, the build needs to initialize the valkey-glide submodule. The process:

1. git submodule update --init --recursive can fail silently (e.g., safe.directory restrictions when PIE runs as root)
2. The fallback relied on .submodule-commits — a file that is empty in the git repo (only populated during PECL package builds)
3. Result: submodule clones at main HEAD instead of the release-pinned commit
4. If main has breaking Rust FFI changes, the build fails with error[E0308]: mismatched types

This was masked in CI because pre-built Rust caches from the correct commit were shared with PIE's build.

### Fix

When git submodule update fails, the build now extracts the correct commit hash from git ls-tree HEAD valkey-glide (the commit recorded in the git tree at the checked-out tag/commit) 
and clones the submodule at that exact commit.

### Changes

- **config.m4**: Added fallback using git ls-tree HEAD in both the early submodule init and the full header generation block. Also fixed PECL path to checkout the commit from 
.submodule-commits after cloning.
- **Makefile.frag**: ensure-submodules target now falls back to git ls-tree HEAD when git submodule update fails.
- **.github/workflows/setup-php-extension/action.yml**: Added conditional submodule init for test protobuf generation.

### Testing

- Verified locally with a script that git ls-tree HEAD resolves the correct pinned commit and the fallback clone+checkout works.
- CI test-pie-install job passes (builds from PR commit with the fix).
- Bug was confirmed on PR #228 where installing v1.1.0 from Packagist fails with Rust compilation errors due to building against main.

### Checklist


config.m4
```diff
@@ -138,14 +138,34 @@ if test "$PHP_VALKEY_GLIDE" != "no"; then
     if test -f ".gitmodules"; then
       if test -d ".git"; then
         dnl Git repository - use submodules
-        git submodule update --init --recursive >/dev/null 2>&1 || true
+        if ! git submodule update --init --recursive >/dev/null 2>&1; then
+          AC_MSG_RESULT([git submodule update failed, falling back to manual clone])
+          if test ! -d "valkey-glide/.git"; then
```
affonsov: this logic is almost repeated here and makefile.frag. can this be extract in way that cover every place that we need?

config.m4
```diff
@@ -138,14 +138,34 @@ if test "$PHP_VALKEY_GLIDE" != "no"; then
...
       else
         dnl build - clone submodules manually during configure
         if test ! -d "valkey-glide/.git"; then
           AC_MSG_RESULT([cloning submodules for build])
           url=$(grep -A1 "path = valkey-glide" .gitmodules | grep url | sed 's/^[[^=]]*=[[[:space:]]]*//')
           if test -n "$url"; then
             git clone "$url" valkey-glide >/dev/null 2>&1 || true
+            dnl Try to checkout the correct commit from .submodule-commits
+            if test -f ".submodule-commits"; then
+              SUBMODULE_COMMIT=$(grep "^valkey-glide=" .submodule-commits | cut -d= -f2)
```
affonsov: roughly same logic as above

config.m4
```diff
@@ -354,14 +374,26 @@ if test "$PHP_VALKEY_GLIDE" != "no"; then
...
-      SUBMODULE_LINE=$(cat .submodule-commits | head -1)
-      SUBMODULE_COMMIT=$(echo "$SUBMODULE_LINE" | cut -d'=' -f2)
-      AC_MSG_RESULT([Debug: submodule line=$SUBMODULE_LINE])
-      AC_MSG_RESULT([Debug: parsed commit=$SUBMODULE_COMMIT])
+    if test ! -d "valkey-glide/.git"; then
+      dnl Determine the correct submodule commit
+      SUBMODULE_COMMIT=""
+      if test -d ".git"; then
+        dnl PIE/git build - get commit from git tree (most reliable)
+        SUBMODULE_COMMIT=$(git ls-tree HEAD valkey-glide 2>/dev/null | awk '{print $3}')
```
affonsov: submodule-commit only exists on the PECL workflow/tar.gz it is not applicable for PIE. We should have a generic logic that applies for PIE and PECL at the same time

config.m4
```diff
@@ -370,14 +402,13 @@ if test "$PHP_VALKEY_GLIDE" != "no"; then
       fi
       
       dnl Clone the submodule at the specific commit
-      git clone --depth 1 https://github.com/valkey-io/valkey-glide.git valkey-glide || AC_MSG_ERROR([Failed to clone valkey-glide])
+      git clone https://github.com/valkey-io/valkey-glide.git valkey-glide || AC_MSG_ERROR([Failed to clone valkey-glide])
```
affonsov: removing dpth 1 is a big issue we don't need to pull all the repo history for the php compilation. Is there any other way to solve this?

.github/workflows/test-pie.yml
```diff
@@ -412,3 +412,44 @@ jobs:
           if [ -f server.pid ]; then
             kill $(cat server.pid) || true
           fi
+
+  # This job installs the released version from Packagist to verify real user experience.
+  # Run manually via "Actions > Run workflow" AFTER the version is published to Packagist.
+  # See: https://bitquill.atlassian.net/wiki/spaces/AMZ/pages/4428562450/PHP+release+process
+  test-pie-install-from-packagist:
```
affonsov, line 351: This will run every time on the PR and it should be run only when we update the packagist

.github/workflows/test-pie.yml
affonsov, line 62: removing this should make the test-pie be similar to the pie installation. This is silently making the test-pie.yml work every time

affonsov (changes requested):

Makefile.frag
```diff
@@ -91,7 +91,21 @@ ensure-submodules:
 	@if [ -f .gitmodules ]; then \
 		if [ -d .git ]; then \
 			echo "Git repository detected - using submodules"; \
-			git submodule update --init --recursive; \
+			if ! git submodule update --init --recursive 2>/dev/null; then \
```
affonsov, line 94: repeated logic

config.m4
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/229/changes/cd70a79ad3caf2007a9e10fa4750fb0ddc00c40d

config.m4
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/229/changes/cd70a79ad3caf2007a9e10fa4750fb0ddc00c40d

config.m4
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/229/changes/cd70a79ad3caf2007a9e10fa4750fb0ddc00c40d

config.m4
prateek-kumar-improving: `--depth 1 clone + fetch --depth 1 origin <sha> `
Using this combination of command works.
Updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/cd70a79ad3caf2007a9e10fa4750fb0ddc00c40d

.github/workflows/test-pie.yml
prateek-kumar-improving: We have a `if: github.event_name == 'workflow_dispatch'` in this.

Makefile.frag
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/cd70a79ad3caf2007a9e10fa4750fb0ddc00c40d

.github/workflows/test-pie.yml
prateek-kumar-improving: the checkout's `submodules: recursive` is needed for CI actions but doesn't affect PIE's isolated build. The real masking (Rust caches) is removed here in this commit: https://github.com/valkey-io/valkey-glide-php/pull/229/changes/5b24b54061783f5ffc70e211b19f6ad51bd8258b

affonsov: As this workflow is testing pie, and in a pie workflow we don't initialize the submdule before installation. Therefore this is not 1:1 to pie installation. The recursive doesn't make sense for what the workflow is for

affonsov: check other places in the workflow that also is using recursive for checkout

affonsov: We probably should first fix the release branch, I think the issue should pop up there, because we don't want to update the rust-core in the release 1.1. What do you think?

prateek-kumar-improving: Checked how to remove`submodules: recursive`. There are other dependencies that will need to be updated manually if we install them separately. Keeping it as it is for now. We can create a backlog issue to more research on this later after the bug is fixed.

affonsov: if testing locally fixed the issue, it should be ok. Not ideal, but ok. Please create an issue

affonsov (approved):
Conditional approval, need to create a backlog issue

.github/workflows/test-pie.yml
prateek-kumar-improving: Created: https://github.com/valkey-io/valkey-glide-php/issues/230

prateek-kumar-improving:
> Conditional approval, need to create a backlog issue

Issue Created: https://github.com/valkey-io/valkey-glide-php/issues/230

alexr-bq (approved):

Issue #236, opened 2026-06-30
Add commands for compatibility with other GLIDE clients


prateek-kumar-improving:
Support for the following commands was added to the Valkey GLIDE Java, Node, Python, and Go clients in https://github.com/valkey-io/valkey-glide/issues/6010. Implement them for Valkey Glide PHP:

- [x] [MIGRATE](https://valkey.io/commands/migrate/) - https://github.com/valkey-io/valkey-glide-php/pull/276
- [x] [RESET](https://valkey.io/commands/reset/) - https://github.com/valkey-io/valkey-glide-php/pull/273
- [x] [CLIENT PAUSE](https://valkey.io/commands/client-pause/) - https://github.com/valkey-io/valkey-glide-php/pull/274
- [x] [CLIENT UNPAUSE](https://valkey.io/commands/client-unpause/) - https://github.com/valkey-io/valkey-glide-php/pull/274
- [x] [CLIENT CACHING](https://valkey.io/commands/client-caching/) - https://github.com/valkey-io/valkey-glide-php/pull/293
- [x] [CLIENT TRACKING](https://valkey.io/commands/client-tracking/) - https://github.com/valkey-io/valkey-glide-php/pull/293
- [x] [CLIENT TRACKINGINFO](https://valkey.io/commands/client-trackinginfo/) - https://github.com/valkey-io/valkey-glide-php/pull/293
- [x] [MONITOR](https://valkey.io/commands/monitor/) - https://github.com/valkey-io/valkey-glide-php/pull/309
- [x] [SAVE](https://valkey.io/commands/save/) - https://github.com/valkey-io/valkey-glide-php/pull/272
- [x] [BGSAVE](https://valkey.io/commands/bgsave/) - https://github.com/valkey-io/valkey-glide-php/pull/237
- [x] [BGREWRITEAOF](https://valkey.io/commands/bgrewriteaof/) - https://github.com/valkey-io/valkey-glide-php/pull/271
- [x] [REPLICAOF](https://valkey.io/commands/replicaof/) - https://github.com/valkey-io/valkey-glide-php/pull/282
- [x] [FAILOVER](https://valkey.io/commands/failover/) - https://github.com/valkey-io/valkey-glide-php/pull/282
- [x] [LATENCY HISTORY](https://valkey.io/commands/latency-history/) - https://github.com/valkey-io/valkey-glide-php/pull/283
[...]

PR #237, opened 2026-06-30
feat(php): implement BGSAVE, BGSAVE SCHEDULE, and BGSAVE CANCEL commands

prateek-kumar-improving:
### Summary

Implement BGSAVE, BGSAVE SCHEDULE, and BGSAVE CANCEL commands for both standalone and cluster clients, matching PHPRedis API and the implementation in other GLIDE clients (Go, Python,
Node, C#).

### Issue link

This Pull Request is linked to issue (URL): #236
Related issues and pull requests:
- [2.5] Implement New Missing Commands valkey-io/valkey-glide#6010
- feat: Add BGSAVE commands valkey-io/valkey-glide-csharp#436

- Add ValkeyGlide::bgSave(?string $mode = null): bool|string for standalone client
- Add ValkeyGlideCluster::bgSave(mixed $route, ?string $mode = null): bool|string for cluster client
- **PHPRedis compatible**: Returns true on success, false on failure by default
- With OPT_REPLY_LITERAL enabled, returns the actual status string (e.g., "Background saving started")
- Supports three modes:
  - null (default): Initiates a background save immediately
  - "SCHEDULE": Schedules a background save
  - "CANCEL": Aborts in-progress/scheduled saves (Valkey 8.1+)

### Implementation

- execute_bgsave_command() added in valkey_glide_commands.c
- BGSAVE_METHOD_IMPL macro added in valkey_glide_commands_common.h
- Macro invoked for ValkeyGlide in valkey_z_php_methods.c
- Macro invoked for ValkeyGlideCluster in valkey_glide_cluster.c
- BgSave case added to prepare_message_args switch in valkey_glide_core_common.c
- Added Map response type handling in process_core_string_result for cluster multi-node responses
- Stub definitions added to valkey_glide.stub.php and valkey_glide_cluster.stub.php

### Testing

- 3 standalone tests: testBgSave, testBgSaveSchedule, testBgSaveCancel
- 3 cluster tests with route parameter
- BGSAVE CANCEL tests version-gated to Valkey 8.1+
- waitForSaveNotInProgress() helper prevents test conflicts from concurrent saves
- Verified OPT_REPLY_LITERAL returns actual string response

### Checklist

### DIFF
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -418,6 +418,30 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         ];
     }
 
+    /**
+     * Override for cluster - info() requires a route parameter.
+     */
+    protected function waitForSaveNotInProgress()
+    {
+        $route  = 'randomNode';
+        $maxWait = 10; // seconds
+        $start   = time();
+        while (time() - $start < $maxWait) {
+            $info = $this->valkey_glide->info($route, 'persistence');
+            if (is_array($info)) {
+                if (
+                    (!isset($info['rdb_bgsave_in_progress']) || $info['rdb_bgsave_in_progress'] != '1') &&
+                    (!isset($info['aof_rewrite_in_progress']) || $info['aof_rewrite_in_progress'] != '1')
+                ) {
+                    return;
+                }
+            } else {
+                return;
+            }
+            usleep(100000); // 100ms
+        }
+    }
+
     public function testBgSave()
     {
         // Wait for any in-progress save to complete
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -430,7 +430,10 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         // Test with allPrimaries route - returns array of node => bool
         $result = $this->valkey_glide->bgSave('allPrimaries');
         $this->assertIsArray($result);
-        foreach ($result as $nodeResult) {
+        $this->assertGT(0, count($result));
+        foreach ($result as $nodeAddress => $nodeResult) {
+            $this->assertIsString($nodeAddress);
+            $this->assertStringContains(':', $nodeAddress);
             $this->assertTrue($nodeResult);
         }
 
@@ -448,7 +451,10 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         // Test with allPrimaries route - returns array of node => bool
         $result = $this->valkey_glide->bgSave('allPrimaries', 'SCHEDULE');
         $this->assertIsArray($result);
-        foreach ($result as $nodeResult) {
+        $this->assertGT(0, count($result));
+        foreach ($result as $nodeAddress => $nodeResult) {
+            $this->assertIsString($nodeAddress);
+            $this->assertStringContains(':', $nodeAddress);
             $this->assertTrue($nodeResult);
         }
 
@@ -471,7 +477,10 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         // Test with allPrimaries route - returns array of node => bool
         $result = $this->valkey_glide->bgSave('allPrimaries', 'CANCEL');
         $this->assertIsArray($result);
-        foreach ($result as $nodeResult) {
+        $this->assertGT(0, count($result));
+        foreach ($result as $nodeAddress => $nodeResult) {
+            $this->assertIsString($nodeAddress);
+            $this->assertStringContains(':', $nodeAddress);
             $this->assertFalse($nodeResult);
         }
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -424,20 +424,10 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $this->waitFor(function () {
             $info = $this->valkey_glide->info('allPrimaries', 'persistence');
-            if (!is_array($info)) {
-                return true;
-            }
             foreach ($info as $nodeInfo) {
-                if (is_array($nodeInfo)) {
-                    if (($nodeInfo['rdb_bgsave_in_progress'] ?? '0') == '1'
-                        || ($nodeInfo['aof_rewrite_in_progress'] ?? '0') == '1') {
-                        return false;
-                    }
-                } elseif (is_string($nodeInfo)) {
-                    if (strpos($nodeInfo, 'rdb_bgsave_in_progress:1') !== false
-                        || strpos($nodeInfo, 'aof_rewrite_in_progress:1') !== false) {
-                        return false;
-                    }
+                if ($nodeInfo['rdb_bgsave_in_progress'] == '1'
+                    || $nodeInfo['aof_rewrite_in_progress'] == '1') {
+                    return false;
                 }
             }
             return true;
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -422,40 +422,26 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
      */
     protected function waitForSaveNotInProgress()
     {
-        $route   = 'allPrimaries';
-        $maxWait = 10; // seconds
-        $start   = time();
-        while (time() - $start < $maxWait) {
-            $info = $this->valkey_glide->info($route, 'persistence');
-            if (is_array($info)) {
-                $saveInProgress = false;
-                foreach ($info as $nodeInfo) {
-                    if (is_array($nodeInfo)) {
-                        if (
-                            (isset($nodeInfo['rdb_bgsave_in_progress']) && $nodeInfo['rdb_bgsave_in_progress'] == '1') ||
-                            (isset($nodeInfo['aof_rewrite_in_progress']) && $nodeInfo['aof_rewrite_in_progress'] == '1')
-                        ) {
-                            $saveInProgress = true;
-                            break;
-                        }
-                    } elseif (is_string($nodeInfo)) {
-                        if (
-                            strpos($nodeInfo, 'rdb_bgsave_in_progress:1') !== false ||
-                            strpos($nodeInfo, 'aof_rewrite_in_progress:1') !== false
-                        ) {
-                            $saveInProgress = true;
-                            break;
-                        }
+        $this->waitFor(function () {
+            $info = $this->valkey_glide->info('allPrimaries', 'persistence');
+            if (!is_array($info)) {
+                return true;
+            }
+            foreach ($info as $nodeInfo) {
+                if (is_array($nodeInfo)) {
+                    if (($nodeInfo['rdb_bgsave_in_progress'] ?? '0') == '1'
+                        || ($nodeInfo['aof_rewrite_in_progress'] ?? '0') == '1') {
+                        return false;
+                    }
+                } elseif (is_string($nodeInfo)) {
+                    if (strpos($nodeInfo, 'rdb_bgsave_in_progress:1') !== false
+                        || strpos($nodeInfo, 'aof_rewrite_in_progress:1') !== false) {
+                        return false;
                     }
                 }
-                if (!$saveInProgress) {
-                    return;
-                }
-            } else {
-                return;
             }
-            usleep(100000); // 100ms
-        }
+            return true;
+        }, 10, 'Timed out waiting for background save to complete on cluster');
     }
 
     public function testBgSave()
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -406,6 +406,18 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         }
     }
 
+    /**
+     * Valid BGSAVE response strings for cluster (includes "OK" from aggregated multi-node response).
+     */
+    protected function bgsaveResponses(): array
+    {
+        return [
+            'Background saving started',
+            'Background saving scheduled',
+            'OK',
+        ];
+    }
+
     public function testBgSave()
     {
         // Wait for any in-progress save to complete
@@ -413,10 +425,10 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $route = 'allPrimaries';
 
-        // BGSAVE with route - returns a non-empty status string
+        // BGSAVE with route - returns a valid status string
         $result = $this->valkey_glide->bgSave($route);
         $this->assertIsString($result);
-        $this->assertNotEquals('', $result);
+        $this->assertContains($result, $this->bgsaveResponses());
 
         // Wait for the bgsave to complete before next test
         $this->waitForSaveNotInProgress();
@@ -432,7 +444,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         // BGSAVE SCHEDULE with route
         $result = $this->valkey_glide->bgSave($route, 'SCHEDULE');
         $this->assertIsString($result);
-        $this->assertNotEquals('', $result);
+        $this->assertContains($result, $this->bgsaveResponses());
 
         // Wait for the scheduled bgsave to complete
         $this->waitForSaveNotInProgress();
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -465,10 +465,9 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $route = 'allPrimaries';
 
-        // BGSAVE with route - returns a valid status string
+        // BGSAVE with route - returns true on success (PHPRedis compatible)
         $result = $this->valkey_glide->bgSave($route);
-        $this->assertIsString($result);
-        $this->assertContains($result, $this->bgsaveResponses());
+        $this->assertTrue($result);
 
         // Wait for the bgsave to complete before next test
         $this->waitForSaveNotInProgress();
@@ -481,10 +480,9 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $route = 'allPrimaries';
 
-        // BGSAVE SCHEDULE with route
+        // BGSAVE SCHEDULE with route - returns true on success
         $result = $this->valkey_glide->bgSave($route, 'SCHEDULE');
-        $this->assertIsString($result);
-        $this->assertContains($result, $this->bgsaveResponses());
+        $this->assertTrue($result);
 
         // Wait for the scheduled bgsave to complete
         $this->waitForSaveNotInProgress();
@@ -503,7 +501,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         // Wait for any in-progress save to complete
         $this->waitForSaveNotInProgress();
 
-        // When no save is in progress, BGSAVE CANCEL should return an error (false)
+        // When no save is in progress, BGSAVE CANCEL should return false
         $result = $this->valkey_glide->bgSave($route, 'CANCEL');
         $this->assertFalse($result);
     }
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -448,37 +448,26 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testBgSave()
     {
-        // Wait for any in-progress save to complete
         $this->waitForSaveNotInProgress();
 
         $route = 'allPrimaries';
 
-        // BGSAVE with route - returns true on success (PHPRedis compatible)
         $result = $this->valkey_glide->bgSave($route);
         $this->assertTrue($result);
-
-        // Wait for the bgsave to complete before next test
-        $this->waitForSaveNotInProgress();
     }
 
     public function testBgSaveSchedule()
     {
-        // Wait for any in-progress save to complete
         $this->waitForSaveNotInProgress();
 
         $route = 'allPrimaries';
 
-        // BGSAVE SCHEDULE with route - returns true on success
         $result = $this->valkey_glide->bgSave($route, 'SCHEDULE');
         $this->assertTrue($result);
-
-        // Wait for the scheduled bgsave to complete
-        $this->waitForSaveNotInProgress();
     }
 
     public function testBgSaveCancel()
     {
-        // BGSAVE CANCEL requires Valkey 8.1+
         if (!$this->minVersionCheck('8.1.0')) {
             $this->markTestSkipped('BGSAVE CANCEL requires Valkey 8.1.0+');
             return;
@@ -486,10 +475,8 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $route = 'allPrimaries';
 
-        // Wait for any in-progress save to complete
         $this->waitForSaveNotInProgress();
 
-        // When no save is in progress, BGSAVE CANCEL should return false
         $result = $this->valkey_glide->bgSave($route, 'CANCEL');
         $this->assertFalse($result);
     }
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -418,20 +418,32 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     }
 
     /**
+     * Override for cluster - info() requires a route parameter and checks all primaries.
+     */
+    protected function isSaveInProgress(): bool
+    {
+        $info = $this->valkey_glide->info('allPrimaries', 'persistence');
+        foreach ($info as $nodeInfo) {
+            if (
+                $nodeInfo['rdb_bgsave_in_progress'] == '1'
+                || $nodeInfo['aof_rewrite_in_progress'] == '1'
+            ) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Override for cluster - info() requires a route parameter.
      */
     protected function waitForSaveNotInProgress()
     {
-        $this->waitFor(function () {
-            $info = $this->valkey_glide->info('allPrimaries', 'persistence');
-            foreach ($info as $nodeInfo) {
-                if ($nodeInfo['rdb_bgsave_in_progress'] == '1'
-                    || $nodeInfo['aof_rewrite_in_progress'] == '1') {
-                    return false;
-                }
-            }
-            return true;
-        }, 10, 'Timed out waiting for background save to complete on cluster');
+        $this->waitFor(
+            fn() => !$this->isSaveInProgress(),
+            10,
+            'Timed out waiting for background save to complete on cluster'
+        );
     }
 
     public function testBgSave()
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -423,18 +423,6 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         return false;
     }
 
-    /**
-     * Override for cluster - info() requires a route parameter.
-     */
-    protected function waitForSaveNotInProgress()
-    {
-        $this->waitFor(
-            fn() => !$this->isSaveInProgress(),
-            10,
-            'Timed out waiting for background save to complete on cluster'
-        );
-    }
-
     public function testBgSave()
     {
         $this->waitForSaveNotInProgress();
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -474,7 +474,14 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $this->waitForSaveNotInProgress();
 
-        // When no save is in progress, CANCEL returns false (error response)
+        // When no save is in progress, CANCEL returns false (not an array) even for
+        // multi-node routes. This is because the glide-core's response aggregation
+        // checks all node responses for errors before building the Map result. When
+        // ALL nodes return a ServerError (no save in progress), the core collapses
+        // them into a single command_error rather than returning a per-node Map.
+        // This matches PHPRedis semantics where bgSave() returns false on failure.
+        // A successful CANCEL (when a save IS in progress) would return an array
+        // for multi-node routes since at least one node returns a success response.
         $result = $this->valkey_glide->bgSave('allPrimaries', 'CANCEL');
         $this->assertFalse($result);
 
@@ -523,6 +530,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         if ($this->minVersionCheck('8.1.0')) {
             $this->waitForSaveNotInProgress();
 
+            // CANCEL with no save in progress returns false (see testBgSaveCancel comment)
             $result = $this->valkey_glide->bgSave('allPrimaries', 'CANCEL');
             $this->assertFalse($result);
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -407,14 +407,13 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     }
 
     /**
-     * Valid BGSAVE response strings for cluster (includes "OK" from aggregated multi-node response).
+     * Valid BGSAVE response strings for cluster.
      */
     protected function bgsaveResponses(): array
     {
         return [
             'Background saving started',
             'Background saving scheduled',
-            'OK',
         ];
     }
 
@@ -423,16 +422,33 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
      */
     protected function waitForSaveNotInProgress()
     {
-        $route  = 'randomNode';
+        $route   = 'allPrimaries';
         $maxWait = 10; // seconds
         $start   = time();
         while (time() - $start < $maxWait) {
             $info = $this->valkey_glide->info($route, 'persistence');
             if (is_array($info)) {
-                if (
-                    (!isset($info['rdb_bgsave_in_progress']) || $info['rdb_bgsave_in_progress'] != '1') &&
-                    (!isset($info['aof_rewrite_in_progress']) || $info['aof_rewrite_in_progress'] != '1')
-                ) {
+                $saveInProgress = false;
+                foreach ($info as $nodeInfo) {
+                    if (is_array($nodeInfo)) {
+                        if (
+                            (isset($nodeInfo['rdb_bgsave_in_progress']) && $nodeInfo['rdb_bgsave_in_progress'] == '1') ||
+                            (isset($nodeInfo['aof_rewrite_in_progress']) && $nodeInfo['aof_rewrite_in_progress'] == '1')
+                        ) {
+                            $saveInProgress = true;
+                            break;
+                        }
+                    } elseif (is_string($nodeInfo)) {
+                        if (
+                            strpos($nodeInfo, 'rdb_bgsave_in_progress:1') !== false ||
+                            strpos($nodeInfo, 'aof_rewrite_in_progress:1') !== false
+                        ) {
+                            $saveInProgress = true;
+                            break;
+                        }
+                    }
+                }
+                if (!$saveInProgress) {
                     return;
                 }
             } else {
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -474,17 +474,10 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $this->waitForSaveNotInProgress();
 
-        // Test with allPrimaries route - returns array of node => bool
+        // When no save is in progress, CANCEL returns false (error response)
         $result = $this->valkey_glide->bgSave('allPrimaries', 'CANCEL');
-        $this->assertIsArray($result);
-        $this->assertGT(0, count($result));
-        foreach ($result as $nodeAddress => $nodeResult) {
-            $this->assertIsString($nodeAddress);
-            $this->assertStringContains(':', $nodeAddress);
-            $this->assertFalse($nodeResult);
-        }
+        $this->assertFalse($result);
 
-        // Test with randomNode route - returns scalar bool
         $result = $this->valkey_glide->bgSave('randomNode', 'CANCEL');
         $this->assertFalse($result);
     }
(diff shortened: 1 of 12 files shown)


valkey_glide.stub.php
```diff
@@ -1153,6 +1153,22 @@ public function flushAll(?bool $sync = null): ValkeyGlide|bool;
...
+     * When called with "CANCEL", aborts all in-progress and scheduled background saves (Valkey 8.1+).
+     *
+     * @param  string|null  $mode  Optional mode: null for default, "SCHEDULE" to schedule,
+     *                             or "CANCEL" to abort (requires Valkey 8.1+).
+     * @return ValkeyGlide|bool|string  Returns true on success, false on failure.
+     *                                  With OPT_REPLY_LITERAL enabled, returns the status string instead.
+     *
+     * @see https://valkey.io/commands/bgsave
+     */
+    public function bgSave(?string $mode = null): ValkeyGlide|bool|string;
```
prateek-kumar-improving, line 1170: Command response type differ when compared to other clients because of client compatibility with PHPRedis.

currantw: Thanks for the explanation!

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+        return [
+            'Background saving started',
+            'Background saving scheduled',
+        ];
+    }
+
+    /**
+     * Override for cluster - info() requires a route parameter.
+     */
+    protected function waitForSaveNotInProgress()
```
currantw: Similar to the other clients, I think it would be helpful to extract a `WaitFor` helper method that generalizes this retry logic so it can be used for other tests.

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+    {
+        $route   = 'allPrimaries';
+        $maxWait = 10; // seconds
+        $start   = time();
+        while (time() - $start < $maxWait) {
+            $info = $this->valkey_glide->info($route, 'persistence');
+            if (is_array($info)) {
+                $saveInProgress = false;
+                foreach ($info as $nodeInfo) {
+                    if (is_array($nodeInfo)) {
```
currantw: Under what conditions might `$info` or `$nodeInfo` not be arrays? If this is not an expected failure, I think we should ignore it to simplify this test helper? Similar with the `isset` checks below.

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+                            strpos($nodeInfo, 'aof_rewrite_in_progress:1') !== false
+                        ) {
+                            $saveInProgress = true;
+                            break;
+                        }
+                    }
+                }
+                if (!$saveInProgress) {
+                    return;
+                }
```
currantw: Would be useful to extract an `isSaveInProgress` helper?

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+        $this->waitForSaveNotInProgress();
+
+        $route = 'allPrimaries';
+
+        // BGSAVE with route - returns true on success (PHPRedis compatible)
+        $result = $this->valkey_glide->bgSave($route);
+        $this->assertTrue($result);
+
+        // Wait for the bgsave to complete before next test
+        $this->waitForSaveNotInProgress();
```
currantw: Is this necessary? We are already calling it at the _start_ of each test, right?

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+    /**
+     * Valid BGSAVE response strings for cluster.
+     */
+    protected function bgsaveResponses(): array
+    {
+        return [
+            'Background saving started',
+            'Background saving scheduled',
+        ];
+    }
```
currantw: 1. Where is this used?
2. This is duplicated. Can we move to shared location?

tests/ValkeyGlideTest.php
```diff
@@ -2585,6 +2585,82 @@ public function testFlushAll()
...
+                    (!isset($info['aof_rewrite_in_progress']) || $info['aof_rewrite_in_progress'] != '1')
+                ) {
+                    return;
+                }
+            } else {
+                return;
+            }
+            usleep(100000); // 100ms
+        }
+    }
```
currantw, line 2705: This logic is mostly duplicated. Is there a way to reduce this?

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+                    return;
+                }
+            } else {
+                return;
+            }
+            usleep(100000); // 100ms
+        }
+    }
+
+    public function testBgSave()
```
currantw, line 426: I think we should match the test cases that are assessed for the other clients. Moreover, given the `OPT_REPLY_LITERAL`, I think there are actually _more_ cases to test for PHP.

Tested for other clients:
- `BGSAVE` standalone + cluster (default route) + cluster (random route)
- `BGSAVE SCHEDULE` standalone + cluster (default route) + cluster (random route)
- `BGSAVE CANCEL` standalone + cluster (default route) + cluster (random route)

In this case, I think we also need to replicate the same tests for `OPT_REPLY_LITERAL` and verify that one of the expected results is returned?

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+        $result = $this->valkey_glide->bgSave($route);
+        $this->assertTrue($result);
+
+        // Wait for the bgsave to complete before next test
+        $this->waitForSaveNotInProgress();
+    }
+
+    public function testBgSaveSchedule()
+    {
+        // Wait for any in-progress save to complete
```
currantw: Nit. Probably don't need this comment. It just duplicates what the method name already tells you 🤷

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,106 @@ public function testFlushAll()
...
+        $result = $this->valkey_glide->bgSave($route, 'SCHEDULE');
+        $this->assertTrue($result);
+
+        // Wait for the scheduled bgsave to complete
+        $this->waitForSaveNotInProgress();
+    }
+
+    public function testBgSaveCancel()
+    {
+        // BGSAVE CANCEL requires Valkey 8.1+
```
currantw: Nit. Similarly, this comment seems to just duplicate what the code says?

valkey_glide_commands.c
```diff
@@ -258,6 +258,86 @@ int execute_flushall_command(zval* object, int argc, zval* return_value, zend_cl
...
+    }
+
+    /* Execute using unified core framework */
+    if (execute_core_command(
+            valkey_glide, &core_args, NULL, process_core_string_result, return_value)) {
+        if (valkey_glide->is_in_batch_mode) {
+            /* In batch mode, return $this for method chaining */
+            ZVAL_COPY(return_value, object);
+            return 1;
+        }
```
currantw: Do we support batch mode for this command?

currantw (commented):
Reviewed to 215e710.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Added WaitFor in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/457659dd0c3e8e031e8e0a821de48644e08f3a6c

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Fixed the code in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/87383a1edb58de7bd7d5ab30ed0233c1fa3a97f3

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Created a helper method in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/5c8bee278a2aff1d1502a7881f33cd032222794a

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated. Removed at the end of all the tests in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/7339acd16a04eca570df512581957d79acb9b548

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Moving to the shared location. It will be required when we add OPT_REPLY_LITERAL tests. Code change commit: https://github.com/valkey-io/valkey-glide-php/pull/237/commits

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Code change commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/6124361aec4b8b227527b726106ac5a8c0ed4d05

Added comprehensive test coverage:
- **Standalone:** testBgSave, testBgSaveSchedule, testBgSaveCancel, testBgSaveWithReplyLiteral
- **Cluster:** Each test now exercises both allPrimaries and randomNode routes, plus testBgSaveWithReplyLiteral tests literal mode with both routes and SCHEDULE

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Removed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/7339acd

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Removed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/7339acd

tests/ValkeyGlideTest.php
prateek-kumar-improving: Removed in previous commits, when helpers were created.

valkey_glide_commands.c
prateek-kumar-improving: We do not explicitly restrict commands in batch mode. User can use it. There is no restriction on valkey side as well.

tests/ValkeyGlideClusterTest.php
currantw: Thanks 🙏

tests/ValkeyGlideClusterTest.php
currantw: Thank you 👍

valkey_glide_commands.c
currantw: Do we have any existing batch integration tests? If so, it would make sense to add tests for these 3 commands in batch. If not, should we add an issue to raise them for all existing tests? This seems like a big gap in test coverage, no?

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,111 @@ public function testFlushAll()
...
+        $result = $this->valkey_glide->bgSave('randomNode');
+        $this->assertIsString($result);
+        $this->assertContains($result, $this->bgsaveResponses());
+
+        $this->waitForSaveNotInProgress();
+
+        // Test SCHEDULE with allPrimaries route
+        $result = $this->valkey_glide->bgSave('allPrimaries', 'SCHEDULE');
+        $this->assertIsString($result);
+        $this->assertContains($result, $this->bgsaveResponses());
```
currantw: Can we also test the `CANCEL` case with `OPT_REPLY_LITERAL` please?

tests/ValkeyGlideTest.php
```diff
@@ -2585,6 +2585,87 @@ public function testFlushAll()
...
+
+        $result = $this->valkey_glide->bgSave();
+        $this->assertIsString($result);
+        $this->assertContains($result, $this->bgsaveResponses());
+
+        $this->waitForSaveNotInProgress();
+
+        $result = $this->valkey_glide->bgSave('SCHEDULE');
+        $this->assertIsString($result);
+        $this->assertContains($result, $this->bgsaveResponses());
```
currantw, line 2631: Similar to previous comment, can we also test `CANCEL` here?

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,111 @@ public function testFlushAll()
...
+     * Override for cluster - info() requires a route parameter.
+     */
+    protected function waitForSaveNotInProgress()
+    {
+        $this->waitFor(
+            fn() => !$this->isSaveInProgress(),
+            10,
+            'Timed out waiting for background save to complete on cluster'
+        );
+    }
```
currantw: Why do we need to override this method?

valkey_glide_core_common.c
```diff
@@ -1590,6 +1591,32 @@ int process_core_string_result(CommandResponse* response, void* output, zval* re
         }
 
         return 1;
+    } else if (response->response_type == Ok) {
+        ZVAL_STRING(return_value, "OK");
+        return 1;
```
currantw, line 1596: Why did we need to add this for this PR specifically? Were there really no other methods previously that returned OK from here?

valkey_glide_core_common.c
```diff
@@ -1590,6 +1591,32 @@ int process_core_string_result(CommandResponse* response, void* output, zval* re
...
+    } else if (response->response_type == Map && response->array_value &&
+               response->array_value_len > 0) {
+        /* Multi-node response (cluster routing) - return first string value */
+        CommandResponse* element     = &response->array_value[0];
+        CommandResponse* first_value = element->map_value;
+        if (first_value && first_value->response_type == String &&
```
currantw: Does this match the behaviour of PHPRedis – does it also only take the first value (even though it is entirely possible that different nodes may return different values)?

valkey_glide_core_common.c
```diff
@@ -1590,6 +1591,32 @@ int process_core_string_result(CommandResponse* response, void* output, zval* re
...
+            first_value->string_value_len > 0) {
+            result = emalloc(first_value->string_value_len + 1);
+            if (!result) {
+                ZVAL_NULL(return_value);
+                return 0;
+            }
+            memcpy(result, first_value->string_value, first_value->string_value_len);
+            result[first_value->string_value_len] = '\0';
+            ZVAL_STRINGL(return_value, result, first_value->string_value_len);
+            efree(result);
```
currantw: This strinc copying logic seems to be duplicated in a few places:

1. Can we extract a helper method?
2. A few lines above, we call `VALKEY_LOG_ERROR` when `emalloc` fails. Should we also do that here?

valkey_glide_core_common.c
```diff
@@ -1590,6 +1591,32 @@ int process_core_string_result(CommandResponse* response, void* output, zval* re
...
+            return 1;
+        } else if (first_value && first_value->response_type == Ok) {
+            ZVAL_STRING(return_value, "OK");
+            return 1;
+        }
+        ZVAL_NULL(return_value);
+        return 0;
```
currantw: This seems to be duplicating logic for handling string and OK return types. Can we extract any common logic so we only need to handle that in one place? Seems like we could just route the first value from a cluster response back to this same method and handle it recursively?

currantw (commented):
Reviewed to 6124361.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/a4a3484fe08d0d57e630b3b48f792ff986dd8648

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/a4a3484fe08d0d57e630b3b48f792ff986dd8648

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Checked this. We can remove this. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/fd083ba72659bc8ab53bf8548ec2fd17e5a046d6

valkey_glide_core_common.c
prateek-kumar-improving: Before this PR, process_core_string_result only handled String and Null response types. 

Commands like SET, FLUSHALL etc. that return "OK" use a different processor — process_core_ok_result (
which just returns true).
We cannot re-use the process_core_ok_result for this because this command as different internal logic.

valkey_glide_core_common.c
prateek-kumar-improving: Code fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/9f9d97603d153af6931cce56b9260815714d2e94

valkey_glide_core_common.c
prateek-kumar-improving: Code fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/9f9d97603d153af6931cce56b9260815714d2e94

valkey_glide_core_common.c
prateek-kumar-improving: Yes, this matches PHPRedis behavior. PHPRedis's bgSave() returns a single bool or string — it doesn't return an array of per-node results.

valkey_glide_commands.c
prateek-kumar-improving: Yes, batch integration tests exist (using multi()/pipeline()/exec()). However, they don't test every command in batch mode — they test specific scenarios with some basic commands. Created a new issue for this: https://github.com/valkey-io/valkey-glide-php/issues/242

currantw: If there is existing infrastructure to support batch command integration tests (which it sounds like there is), then I think we should add them as part of this issue. One test per commands (i.e. `BGSAVE`, `BGSAVE SCHEDULE`, and `BGSAVE CANCEL` should suffice.

Can you please also add additional information to that pull request? Should be in a state where a future developer could pick it up and start on it – including someone not on our team. Information I would include:

- What support already exists for batch command tests?
- What commands are already tested? What commands are _not_ tested?
- What are some possible divisions of work? i.e. how could this be broken down by command group.

valkey_glide_core_common.c
currantw: > We cannot re-use the process_core_ok_result for this because this command as different internal logic.

Can you explain a bit more what the "different internal logic" is here that means that we can't re-use any of the `process_core_string_result` logic?

valkey_glide_core_common.c
currantw: Did you confirm (either by running it, from the documentation, or from the code) that the value returned by PHPRedis is simply the value from one of the nodes? Can you explain how you verified this?

currantw (commented):
Reviewed to 9f9d976.

valkey_glide_core_common.c
prateek-kumar-improving: We are re-using `process_core_string_result`.
What I meant was the existing process_core_ok_result will not work here.

valkey_glide_core_common.c
prateek-kumar-improving: Verified from PHPRedis documentation ([cluster.md](https://github.com/phpredis/phpredis/blob/develop/cluster.md)). BGSAVE is listed as a "Directed node command" — in RedisCluster, it's sent to a single node (specified by the caller) and returns that node's scalar response. 

PHPRedis does not aggregate multi-node results. Our allPrimaries route is a GLIDE-specific extension; taking the first value maintains the scalar return type that PHPRedis users expect.

When a user passes a key or host:port, they get a single-node response (just like PHPRedis). The "take first value from Map" logic only exists for multi-node routes (allPrimaries, allNodes). 
PHPRedis doesn't have an equivalent to allPrimaries — it's a GLIDE addition.

valkey_glide_commands.c
prateek-kumar-improving: Added Batch tests for BGSAVE: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/258411d00a30695e613cd340a2516faba8f06ef5

valkey_glide_core_common.c
```diff
@@ -1590,6 +1591,19 @@ int process_core_string_result(CommandResponse* response, void* output, zval* re
...
+        ZVAL_STRING(return_value, "OK");
+        return 1;
+    } else if (response->response_type == Map && response->array_value &&
+               response->array_value_len > 0) {
+        /* Multi-node response (cluster routing) - process first node's value recursively */
+        CommandResponse* element     = &response->array_value[0];
+        CommandResponse* first_value = element->map_value;
+        if (first_value) {
+            return process_core_string_result(first_value, output, return_value);
+        }
```
Aryex: Why is this only looking at the first `array_value`? Should we also process the other responses as well?

prateek-kumar-improving: PHPRedis bgSave() returns a single scalar — it doesn't aggregate multi-node results. In PHPRedis cluster, BGSAVE is a "directed node command" sent to one specific node. Our allPrimaries route is a GLIDE-specific extension; we take the first value to maintain the scalar return type.

Aryex: Would this correctly handles `bgSave("allPrimaries")`?

valkey_glide_commands.c
currantw: Awesome, thank you 🚀

valkey_glide_core_common.c
currantw: In that case, should GLIDE return return a dictionary of strings (map from address to response string) for multi-node routes? If multi-node routes are not supported by PHPRedis at all, then this would be an _extension_ of the PHPRedis API, rather than a change to it (i.e. the result would still be compatible for PHPRedis-support commands), _**and**_ it would be consistent with the other GLIDE clients (which return a dictionary/ClusterValue for multi-node routes). If we do not return dictionary for multi-node commands, than we hiding responses from the user and mis-representing the actual responses from the server.

valkey_glide_core_common.c
currantw: I think that we may need to return a dictionary/ClusterValue for multi-node routes, as suggested in this comment: https://github.com/valkey-io/valkey-glide-php/pull/237#discussion_r3532728888. This is what other GLIDE clients do, and I believe that it is still compatible with PHPRedis behaviour.

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,142 @@ public function testFlushAll()
...
+        $result = $this->valkey_glide->bgSave('randomNode');
+        $this->assertIsString($result);
+        $this->assertContains($result, $this->bgsaveResponses());
+
+        $this->waitForSaveNotInProgress();
+
+        // Test SCHEDULE with allPrimaries route
+        $result = $this->valkey_glide->bgSave('allPrimaries', 'SCHEDULE');
+        $this->assertIsString($result);
+        $this->assertContains($result, $this->bgsaveResponses());
```
currantw: Seems like we are missing a `randomNode` route for `SCHEDULE`?

currantw (commented):
Reviewed to 4302cf1.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/7732a987886188971cd74b14567d4073b5d81f64

valkey_glide_core_common.c
prateek-kumar-improving: Extended the implementation by adding an array return type: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/646893a94c03e883814149d8a229dcbdec59a910, https://github.com/valkey-io/valkey-glide-php/pull/237/changes/5d20fcb7bdec697672de00bd93cde188e7a10b4b

valkey_glide_core_common.c
prateek-kumar-improving: Extended the implementation by adding an array return type: https://github.com/valkey-io/valkey-glide-php/commit/646893a94c03e883814149d8a229dcbdec59a910, https://github.com/valkey-io/valkey-glide-php/commit/5d20fcb7bdec697672de00bd93cde188e7a10b4b

Aryex (approved):

valkey_glide_cluster.stub.php
```diff
@@ -558,6 +558,11 @@ public function flushAll(mixed $route, bool $async = false): ValkeyGlideCluster|
      */
     public function flushDB(mixed $route, bool $async = false): ValkeyGlideCluster|bool;
 
+    /**
+     * @see ValkeyGlide::bgSave
```
currantw, line 562: The differences in return type and behaviour between standalone and cluster are not documented here.

tests/ValkeyGlideClusterTest.php
```diff
@@ -406,6 +406,170 @@ public function testFlushAll()
...
+        if (!$this->minVersionCheck('8.1.0')) {
+            $this->markTestSkipped('BGSAVE CANCEL requires Valkey 8.1.0+');
+            return;
+        }
+
+        $this->waitForSaveNotInProgress();
+
+        // When no save is in progress, CANCEL returns false (error response)
+        $result = $this->valkey_glide->bgSave('allPrimaries', 'CANCEL');
+        $this->assertFalse($result);
```
currantw: Why does `BGSAVE CANCEL` not return an `array` like the other methods? This command could still succeed on some nodes and fail on others, right? 

For other clients, failure causes an exception to be thrown, so we are not losing information if we return "OK"; but in this case, where – as I understand it – failure would NOT cause an exception, we are losing information if we don't return a bool. Moreover, it is unclear what this bool actually means. Does `true` mean that ALL nodes returned true? Or just one? Or most of them?

valkey_glide_commands.c
```diff
@@ -258,6 +258,86 @@ int execute_flushall_command(zval* object, int argc, zval* return_value, zend_cl
...
+    /* Add mode option if provided (SCHEDULE or CANCEL) */
+    if (mode && mode_len > 0) {
+        core_args.args[0].type                  = CORE_ARG_TYPE_STRING;
+        core_args.args[0].data.string_arg.value = mode;
+        core_args.args[0].data.string_arg.len   = mode_len;
+        core_args.arg_count                     = 1;
+    }
+
+    /* Select processor based on OPT_REPLY_LITERAL:
+     * - With OPT_REPLY_LITERAL: return raw string (process_core_string_result)
```
currantw: Should this be `process_core_bgsave_string_result` instead?


```suggestion
     * - With OPT_REPLY_LITERAL: return raw string (process_core_bgsave_string_result)
```

valkey_glide_commands.c
```diff
@@ -258,6 +258,86 @@ int execute_flushall_command(zval* object, int argc, zval* return_value, zend_cl
...
+        if (valkey_glide->is_in_batch_mode) {
+            /* In batch mode, return $this for method chaining */
+            ZVAL_COPY(return_value, object);
+            return 1;
+        }
+
+        return 1;
+    } else {
+        return 0;
+    }
```
currantw: Nit. Can simplify this. Reduces return paths/branching and nesting:


```suggestion
    if (!execute_core_command(valkey_glide, &core_args, NULL, processor, return_value)) {
        return 0;
    }
        
    if (valkey_glide->is_in_batch_mode) {
        /* In batch mode, return $this for method chaining */
        ZVAL_COPY(return_value, object);
    }

    return 1;
```

valkey_glide_core_common.c
```diff
@@ -1623,6 +1624,89 @@ int process_core_bool_result(CommandResponse* response, void* output, zval* retu
     return 1;
 }
 
+/**
+ * Batch-compatible wrapper for BGSAVE boolean results.
+ * Converts String/Ok responses to true, Null to false.
+ * For multi-node routes, returns an associative array of node => bool.
+ */
+int process_core_bgsave_bool_result(CommandResponse* response, void* output, zval* return_value) {
```
currantw: I think we can significantly simplify this and reduce duplication by implementing something like `process_core_cluster_result`, which would take a converter function and apply it to a single- or multi-node cluster result. This would be similar to logic that exists in other clients, and would avoid duplication for new commands. 

Here is an AI generated example of what this could look like:

```php
  int process_core_cluster_result(
      CommandResponse* response,
      void* output,
      zval* return_value,
      z_result_processor_t single_node_processor
  ) {
      if (!response) {
          ZVAL_FALSE(return_value);
          return 0;
      }

      if (response->response_type == Map && response->array_value &&
          response->array_value_len > 0) {
          array_init(return_value);
          for (long i = 0; i < response->array_value_len; i++) {
              CommandResponse* element = &response->array_value[i];
              CommandResponse* value   = element->map_value;
              if (element->map_key && element->map_key->string_value && value) {
                  zval node_result;
                  single_node_processor(value, output, &node_result);
                  add_assoc_zval_ex(return_value,
                                    element->map_key->string_value,
                                    element->map_key->string_value_len,
                                    &node_result);
              }
          }
          return 1;
      }

      // Single-node: delegate directly
      return single_node_processor(response, output, return_value);
  }

  int process_core_cluster_bool_result(CommandResponse* r, void* o, zval* rv) {
      return process_core_cluster_result(r, o, rv, process_core_bool_result);
  }

  int process_core_cluster_string_result(CommandResponse* r, void* o, zval* rv) {
      return process_core_cluster_result(r, o, rv, process_core_string_result);
```

This is much simpler, and has the advantage of being more generic and reusable.

currantw (commented):
Reviewed to a657785.

valkey_glide_core_common.c
currantw: Thanks! I have reviewed these changes and added a few more suggestions, but generally looks good – I think this is a much improved interface that more closely matches other GLIDE clients.

tests/ValkeyGlideClusterTest.php
currantw: Thanks!

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/7a122c9589adb5a59f4d8ab070607f061ef2d630

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Added a comment for this: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/d2d19919ccd324709bf1c6ca60d2659eee78a9b8.

It's a limitation of the glide-core's response aggregation, not a PHP-level design choice. For BGSAVE (which has no explicit ResponsePolicy in redis-rs), multi-node routing uses the None policy path. Before building the Value::Map of per-node results, the core iterates all resolved values and returns the first ServerError as a single Err if any node errors. When CANCEL is called with no save in progress, ALL nodes return errors, so the core collapses them into a single command_error — we never receive a Map response on the PHP side.

This means:
   - Success case (save was in progress): at least one node returns OK → no error short-circuit → we get the Map → PHP returns array<string, bool>
  - All-failure case (no save in progress): all nodes error → core returns single Err → PHP returns scalar false

This matches PHPRedis semantics (returns false on failure) and is consistent with how the Python GLIDE client handles this — Python would throw a RequestError in the same scenario rather than returning a map.

Changing this would require modifying the core's ResponsePolicy for BGSAVE (e.g., to Special) so per-node errors are preserved in the Map, which would be a separate core-level change, which would require a change for all clients.

valkey_glide_commands.c
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/83da041fc2c66c796c2157aa7d5574ae535e52a8

valkey_glide_commands.c
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/83da041fc2c66c796c2157aa7d5574ae535e52a8

valkey_glide_core_common.c
prateek-kumar-improving: Refactored here: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/d370bcd2c6f6e9e8245f7d800d47f3ff93e02ddc

valkey_glide_cluster.stub.php
```diff
@@ -558,6 +558,40 @@ public function flushAll(mixed $route, bool $async = false): ValkeyGlideCluster|
...
+     *
+     * @param mixed        $route  The routing configuration that determines which node(s) to send the
+     *                             command to. Can be:
+     *                             - string "randomNode" to route to a random node
+     *                             - string "allPrimaries" to route to all primary nodes
+     *                             - string "allNodes" to route to all nodes (primaries and replicas)
+     *                             - string containing a key name for slot-based routing
+     *                             - array ['type' => 'primarySlotKey', 'key' => 'keyName'] for slot key routing
+     *                             - array ['type' => 'routeByAddress', 'host' => 'hostname', 'port' => port]
+     *                               for specific node routing
```
currantw: Nit. How are `$route` commands documented elsewhere? We should match that, I think. Seems like it would be overkill to duplicate documentation for _all_ the route options on each method that accepts a `$route` parameter... 🤷 

Please resolve once addressed.

tests/ValkeyGlideClusterTest.php
currantw: Sounds good. Thanks for the explanation.

valkey_glide_core_common.c
```diff
@@ -1623,6 +1624,115 @@ int process_core_bool_result(CommandResponse* response, void* output, zval* retu
...
+
+    /* Single-node response - delegate directly */
+    return single_node_processor(response, output, return_value);
+}
+
+/**
+ * Single-node BGSAVE bool processor.
+ * Converts String/Ok responses to true, anything else to false.
+ */
+static int process_bgsave_single_node_bool(CommandResponse* response,
```
currantw, line 1671: Why can't we just use `process_core_bool_result` instead of this new method?

valkey_glide_core_common.c
```diff
@@ -1623,6 +1624,115 @@ int process_core_bool_result(CommandResponse* response, void* output, zval* retu
...
+
+    ZVAL_FALSE(return_value);
+    return 1;
+}
+
+/**
+ * Single-node BGSAVE string processor (OPT_REPLY_LITERAL mode).
+ * Returns the status string, or false for Null responses.
+ */
+static int process_bgsave_single_node_string(CommandResponse* response,
```
currantw: Why can't we just use `process_core_string_result` instead of creating a new method?

currantw (commented):
Reviewed to 84db088. Just a couple more comments.

valkey_glide_cluster.stub.php
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/18326af7e581b93d3481cc40d30d44a7361d09e4 and this: https://github.com/valkey-io/valkey-glide-php/pull/237/changes/8b9411b5f35780117ecf7edc3a0d75fe2fd283ea

valkey_glide_core_common.c
prateek-kumar-improving: We can't use process_core_bool_result directly because it handles Bool, Int, and Ok response types — but BGSAVE returns a String status (e.g., "Background saving started") on success. The generic bool 
processor doesn't treat String as true, so it would return false for successful BGSAVE responses.

process_bgsave_single_node_bool specifically maps String || Ok → true and everything else to false, which matches BGSAVE's actual response format. This edge case requires special handling different from the generic case.

valkey_glide_core_common.c
prateek-kumar-improving: We can use process_core_string_result directly here. The only difference was Null response handling (NULL vs false), again an edge case.
The possibility of bgSave returning null is not much but we still don't want it returning null instead of false for some edge case.

currantw: I'm sorry, I'm still a bit confused.

The current implementation of `process_bgsave_single_node_string` handles `Null` this way:

```c
    } else if (response->response_type == Null) {
        ZVAL_FALSE(return_value);
        return 1;
    }
```

Meanwhile, the current implementation of `process_core_string_result` has this:

```c
    } else if (response->response_type == Null) {
        ZVAL_FALSE(return_value);
        return 1;
    }
```

These seems to be identical to me? What am I missing?

The only difference between these two seems to be that `process_bgsave_single_node_string` handles OK as well. It seems to me like it would be reasonable to just update `process_core_string_result` to also handle OK and call it a day?

valkey_glide_core_common.c
currantw: > The generic bool processor doesn't treat String as true, so it would return false for successful BGSAVE responses

Would there by any advantage to it doing so? It seems reasonable to me that a non-empty string would map to true? We already do that same sort of mapping with OK, right?

prateek-kumar-improving: `process_core_bool_result` - It maps the **string type to false** and this method is called 22 times in the code.

If we use `process_core_bool_result` and then make it return true for this case, it would map the string type returned from other methods to **true** instead of **false**.

We just cannot change the handling of string types for 22 other methods. There is a possibility that a lot of the methods don't return string but it is the common base case that returns **false** for those 22 methods, so, best not to change it and introduce regression. Our case is the exception, therefore best to keep it separate and return **true for string**.

There is no advantage in this but we have to keep it separate because of this case. 

We can definitely refactor and make the PHP code better at this layer of the code so that we don't have to introduce regression in already released command, but that again can't be done here. For refactoring in this PR we will have to check every command that uses this method and then check that all its cases and tests are handled properly. Ideally, code should be extendable enough avoid this kind of regression but based on the current design this is the best way to avoid regression and extend the code.

We can refactor this as a separate issue. What do you think?

valkey_glide_core_common.c
prateek-kumar-improving: In this case, I agree it is reasonable to use the same method as it is not changing an existing case's return value. It is just adding Ok to the existing method.

Before the previous refactoring the Null case was handled separately so this change should work.

valkey_glide_core_common.c
currantw: Sounds good! I think this definitely needs a cleanup at this point – but doesn't need to be in this PR! I just didn't want to introduce any unnecessary complexity here that would make this difficult in the future. I think you've accomplished that. Please resolve this once you've read.

valkey_glide_core_common.c
currantw: Thanks!

currantw (approved):
Looks good! Thanks for all your work on this. ✅

PR #272, opened 2026-07-21
Implement save command

prateek-kumar-improving:
### Summary

Implement the SAVE command for both standalone and cluster clients.

### Issue link

- Add ValkeyGlide::save(): ValkeyGlide|bool|string for standalone client
- Add ValkeyGlideCluster::save(mixed $route): ValkeyGlideCluster|array|bool|string for cluster client
- Returns true on success by default
- With OPT_REPLY_LITERAL enabled, returns the raw status string "OK"
- Cluster multi-node routes return an associative array of node => bool|string

### Implementation

- execute_save_command() in valkey_glide_commands.c — zero-arg command using parse_cluster_route() helper and execute_and_handle_batch() helper
- SAVE_METHOD_IMPL macro in valkey_glide_commands_common.h
- Save case added to prepare_zero_args switch in valkey_glide_core_common.c
- Reuses existing process_core_status_bool_result / process_core_status_string_result processors
- **Refactoring**: execute_and_handle_batch() helper extracted and applied to flushdb, flushall, bgsave, bgrewriteaof, time, pfadd, pfmerge (net -58 lines)

### Limitations

None. Full feature parity with other GLIDE clients.

### Testing

1. Standalone tests (tests/ValkeyGlideTest.php) updated.
2. Cluster tests (tests/ValkeyGlideClusterTest.php) updated.
3. All tests use waitForSaveNotInProgress() pre-condition to avoid flakiness.

### Checklist

### DIFF
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -496,38 +496,46 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
                 $this->assertIsString($nodeResult);
                 $this->assertContains($nodeResult, $this->bgsaveResponses());
             }
+        });
 
-            $this->waitForSaveNotInProgress();
+        $this->waitForSaveNotInProgress();
 
+        $this->withOptReplyLiteralEnabled(function () {
             $result = $this->valkey_glide->bgSave('randomNode');
             $this->assertIsString($result);
             $this->assertContains($result, $this->bgsaveResponses());
+        });
 
-            $this->waitForSaveNotInProgress();
+        $this->waitForSaveNotInProgress();
 
+        $this->withOptReplyLiteralEnabled(function () {
             $result = $this->valkey_glide->bgSave('allPrimaries', 'SCHEDULE');
             $this->assertIsArray($result);
             foreach ($result as $nodeResult) {
                 $this->assertIsString($nodeResult);
                 $this->assertContains($nodeResult, $this->bgsaveResponses());
             }
+        });
 
-            $this->waitForSaveNotInProgress();
+        $this->waitForSaveNotInProgress();
 
+        $this->withOptReplyLiteralEnabled(function () {
             $result = $this->valkey_glide->bgSave('randomNode', 'SCHEDULE');
             $this->assertIsString($result);
             $this->assertContains($result, $this->bgsaveResponses());
+        });
 
-            if ($this->minVersionCheck('8.1.0')) {
-                $this->waitForSaveNotInProgress();
+        if ($this->minVersionCheck('8.1.0')) {
+            $this->waitForSaveNotInProgress();
 
+            $this->withOptReplyLiteralEnabled(function () {
                 $result = $this->valkey_glide->bgSave('allPrimaries', 'CANCEL');
                 $this->assertFalse($result);
 
                 $result = $this->valkey_glide->bgSave('randomNode', 'CANCEL');
                 $this->assertFalse($result);
-            }
-        });
+            });
+        }
     }
 
     public function testBgSaveBatch()
@@ -600,9 +608,11 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
                 $this->assertIsString($nodeResult);
                 $this->assertContains($nodeResult, $this->bgRewriteAofResponses());
             }
+        });
 
-            $this->waitForSaveNotInProgress();
+        $this->waitForSaveNotInProgress();
 
+        $this->withOptReplyLiteralEnabled(function () {
             $result = $this->valkey_glide->bgRewriteAof('randomNode');
             $this->assertIsString($result);
             $this->assertContains($result, $this->bgRewriteAofResponses());
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -413,11 +413,20 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $info = $this->valkey_glide->info('allPrimaries', 'persistence');
         foreach ($info as $nodeInfo) {
-            if (
-                $nodeInfo['rdb_bgsave_in_progress'] == '1'
-                || $nodeInfo['aof_rewrite_in_progress'] == '1'
-            ) {
-                return true;
+            if (is_array($nodeInfo)) {
+                if (
+                    (isset($nodeInfo['rdb_bgsave_in_progress']) && $nodeInfo['rdb_bgsave_in_progress'] == '1')
+                    || (isset($nodeInfo['aof_rewrite_in_progress']) && $nodeInfo['aof_rewrite_in_progress'] == '1')
+                ) {
+                    return true;
+                }
+            } elseif (is_string($nodeInfo)) {
+                if (
+                    str_contains($nodeInfo, 'rdb_bgsave_in_progress:1')
+                    || str_contains($nodeInfo, 'aof_rewrite_in_progress:1')
+                ) {
+                    return true;
+                }
             }
         }
         return false;
@@ -648,15 +657,18 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $this->waitForSaveNotInProgress();
 
-        // Test with allPrimaries route - returns array of node => bool
+        // Test with randomNode route - returns scalar bool
+        $result = $this->valkey_glide->save('randomNode');
+        $this->assertTrue($result);
+    }
+
+    public function testSaveWithRoute()
+    {
+        $this->waitForSaveNotInProgress();
+
+        // Test with allPrimaries route - succeeds without error
         $result = $this->valkey_glide->save('allPrimaries');
-        $this->assertIsArray($result);
-        $this->assertGT(0, count($result));
-        foreach ($result as $nodeAddress => $nodeResult) {
-            $this->assertIsString($nodeAddress);
-            $this->assertStringContains(':', $nodeAddress);
-            $this->assertTrue($nodeResult);
-        }
+        $this->assertTrue($result);
     }
 
     public function testSaveWithRoute()
@@ -674,17 +686,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
 
-        // Test with allPrimaries route - returns array of node => string
-        $result = $this->valkey_glide->save('allPrimaries');
-        $this->assertIsArray($result);
-        foreach ($result as $nodeResult) {
-            $this->assertIsString($nodeResult);
-            $this->assertEquals('OK', $nodeResult);
-        }
-
-        $this->waitForSaveNotInProgress();
-
-        // Test with randomNode route - returns scalar string
+        // Test with randomNode route - returns scalar string "OK"
         $result = $this->valkey_glide->save('randomNode');
         $this->assertIsString($result);
         $this->assertEquals('OK', $result);
@@ -702,12 +704,9 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->waitForSaveNotInProgress();
 
         $this->valkey_glide->pipeline();
-        $this->valkey_glide->save('allPrimaries');
+        $this->valkey_glide->save('randomNode');
         $result = $this->valkey_glide->exec();
-        $this->assertIsArray($result[0]);
-        foreach ($result[0] as $nodeResult) {
-            $this->assertTrue($nodeResult);
-        }
+        $this->assertTrue($result[0]);
     }
 
     public function testInfo()
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -493,52 +493,45 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $this->waitForSaveNotInProgress();
 
-        $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
+        $this->withOptReplyLiteralEnabled(function () {
+            $result = $this->valkey_glide->bgSave('allPrimaries');
+            $this->assertIsArray($result);
+            foreach ($result as $nodeResult) {
+                $this->assertIsString($nodeResult);
+                $this->assertContains($nodeResult, $this->bgsaveResponses());
+            }
 
-        // Test with allPrimaries route - returns array of node => string
-        $result = $this->valkey_glide->bgSave('allPrimaries');
-        $this->assertIsArray($result);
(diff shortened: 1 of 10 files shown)


valkey_glide.stub.php
```diff
@@ -1200,6 +1200,19 @@ public function bgSave(?string $mode = null): ValkeyGlide|bool|string;
...
     public function bgRewriteAof(): ValkeyGlide|bool|string;
 
+    /**
+     * Synchronously saves the dataset to disk (blocking).
+     *
+     * This command blocks the server until the save is complete. For a non-blocking
+     * alternative, use {@see bgSave()}.
+     *
+     * @return ValkeyGlide|bool|string  Returns true on success, false on failure.
+     *                                  With OPT_REPLY_LITERAL enabled, returns "OK" instead.
```
currantw: What is the failure behaviour here, esp. when `OPT_REPLY_LITERAL` is enabled?

valkey_glide_cluster.stub.php
```diff
@@ -584,6 +584,13 @@ public function bgSave(mixed $route, ?string $mode = null): ValkeyGlideCluster|a
      */
     public function bgRewriteAof(mixed $route): ValkeyGlideCluster|array|bool|string;
 
+    /**
+     * @see ValkeyGlide::save
+     *
+     * Returns true on success (or "OK" with OPT_REPLY_LITERAL enabled).
```
currantw: This doesn't seem consistent with the other methods above/below?

valkey_glide_commands_common.h
```diff
@@ -721,6 +722,20 @@ int execute_unlink_command(zval* object, int argc, zval* return_value, zend_clas
...
+        if (execute_save_command(getThis(),                                     \
+                                 ZEND_NUM_ARGS(),                               \
+                                 return_value,                                  \
+                                 strcmp(#class_name, "ValkeyGlideCluster") == 0 \
+                                     ? get_valkey_glide_cluster_ce()            \
+                                     : get_valkey_glide_ce())) {                \
+            return;                                                             \
+        }                                                                       \
+        zval_dtor(return_value);                                                \
+        RETURN_FALSE;                                                           \
```
currantw: This block seems to be repeated many times in this file. Can it be extracted to a helper?

valkey_glide_core_common.c
```diff
@@ -203,6 +203,7 @@ int prepare_core_args(core_command_args_t* args,
         case Role:
         case DBSize:
         case BgRewriteAof:
+        case Save:
```
currantw: Nit. When I have blocks like this, I find it is helpful to order them alphabetically. Makes it easier for future devs for quickly see if a value is in the list. Same with other blocks below. 🤷

tests/ValkeyGlideTest.php
```diff
@@ -2729,6 +2729,42 @@ protected function bgRewriteAofResponses(): array
...
+        $this->waitForSaveNotInProgress();
+
+        $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
+
+        $result = $this->valkey_glide->save();
+        $this->assertIsString($result);
+        $this->assertEquals('OK', $result);
+
+        $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, false);
+    }
```
currantw, line 2763: There are two problems that I see with the current test pattern for `OPT_REPLY_LITERAL`:

1. If the method throws an exception (i.e. because the test times out), the `OPT_REPLY_LITERAL` will never be disabled. This could cause a bunch of confusing, difficult-to-debug test failures.
2. These lines of code (enabled and disable) are repeated in every test. This would get worse if we needed to guard everything in try/catch blocks.

Given that, I propose adding a `withOptReplyLiteralEnabled` helper, something like:

```php
private function withOptReplyLiteralEnabled(callable $fn)
{
    $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
    try {
        return $fn();
    } finally {
        $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, false);
    }
}

public function testSaveWithReplyLiteral()
{
    $this->waitForSaveNotInProgress();

    $this->withOptReplyLiteralEnabled(function () {
        $result = $this->valkey_glide->save();

        $this->assertIsString($result);
        $this->assertEquals('OK', $result);
    });
}
```

tests/ValkeyGlideClusterTest.php
```diff
@@ -414,8 +414,8 @@ protected function isSaveInProgress(): bool
         $info = $this->valkey_glide->info('allPrimaries', 'persistence');
         foreach ($info as $nodeInfo) {
             if (
-                $nodeInfo['rdb_bgsave_in_progress'] == '1'
-                || $nodeInfo['aof_rewrite_in_progress'] == '1'
+                (isset($nodeInfo['rdb_bgsave_in_progress']) && $nodeInfo['rdb_bgsave_in_progress'] == '1')
+                || (isset($nodeInfo['aof_rewrite_in_progress']) && $nodeInfo['aof_rewrite_in_progress'] == '1')
```
currantw, line 418: Why was this change needed? Under what circumstances might the node info response not be a set?

tests/ValkeyGlideClusterTest.php
```diff
@@ -644,6 +644,53 @@ public function testBgRewriteAofBatch()
...
+        $this->assertTrue($result);
+    }
+
+    public function testSaveWithReplyLiteral()
+    {
+        $this->waitForSaveNotInProgress();
+
+        $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
+
+        // Test with randomNode route - returns scalar string "OK"
```
currantw: Nit. Similar to previous review comments, I'm not sure what these comments are adding that can't also be cleaned from the code already in roughly the same number of characters? Similar with two methods above. 🤷

currantw (commented):
Reviewed to 554cae0.

valkey_glide_commands.c
```diff
@@ -363,6 +363,48 @@ int execute_bgrewriteaof_command(zval* object, int argc, zval* return_value, zen
...
+
+    /* Select processor based on OPT_REPLY_LITERAL:
+     * - With OPT_REPLY_LITERAL: return raw string (process_core_status_string_result)
+     * - Without OPT_REPLY_LITERAL: return bool (process_core_status_bool_result)
+     * This matches PHPRedis behavior where save() returns bool. */
+    z_result_processor_t processor = valkey_glide->opt_reply_literal
+                                         ? process_core_status_string_result
+                                         : process_core_status_bool_result;
+
+    /* Execute using unified core framework */
```
currantw: Nit. Is this comment necessary? Not sure what additional information it provides? Same with other methods?

valkey_glide.stub.php
prateek-kumar-improving: It returns false. Same for the entire client. Updated the doc here: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/4a3290f4f67ecf50a160eb69fabea273e5cc3211

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/1fa3cdaa30e6b325fd1b9c69d31028560878c0be

valkey_glide_commands_common.h
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/88578488a1d14b95666153501e5dc5f040dcfe21

valkey_glide_core_common.c
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/2d8b286cc0b226b15591e1625634e31811d862c0

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/c517548f1f6f32938ac539f8627a8c7fa4ec509e

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: The isset() guards were added for cluster mode. When info('persistence') is called with a multi-node route, we iterate over each node's parsed info response. In rare cases (topology changes, node in transitional state, or version differences), a node might return a partial info section that doesn't include these specific fields. 

Without isset(), PHP would emit an Undefined array key warning.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/597696fbac26b90d26ce9ed1b37dc54a933c7fc2

tests/ValkeyGlideTest.php
```diff
@@ -2618,26 +2632,24 @@ public function testBgSaveWithReplyLiteral()
...
 
-            $result = $this->valkey_glide->bgSave('CANCEL');
-            $this->assertFalse($result);
-        }
+            $result = $this->valkey_glide->bgSave('SCHEDULE');
+            $this->assertIsString($result);
+            $this->assertContains($result, $this->bgsaveResponses());
+
+            if ($this->minVersionCheck('8.1.0')) {
+                $this->waitForSaveNotInProgress();
```
currantw: Is there any risk to having `waitForSaveNotInProgress` calls inside `withOptReplyLiteralEnabled`? Probably makes sense to be consistent (either have the all inside the methods or all outside).

currantw (commented):
Reviewed to 597696f.

valkey_glide_commands_common.h
currantw: Screenshot 2026-07-22 at 11 23 58 AM

Love to see it! (Feel free to resolve).

tests/ValkeyGlideClusterTest.php
currantw: Can you explain this a bit more? What evidence did you collect for this? Why was this not identified previously? Do you have any test failures you can share that demonstrate this? I haven't observed this in C#, so I'm curious why this would happen in the PHP client?

currantw (commented):
Reviewed to 597696f.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Got this error, the **undefined array** one only once. Most probably was something flaky because of some bug in one of the previous implementations. Kept the check because it is just extra validation.

Was unable to replicate the error after that. Can remove it if you think it is an issue. 
Ideally, I think if for some reason the key is not set because of any issue, it is better to have the failure being thrown later using the assert statement.
This can happen if there are cases like some response changes because of version change, protocol etc.

valkey_glide_commands.c
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/4c4d79d3630e76554076903811e4a392ca9e5490

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/823d2471e7568da113d43511ea2957afca6d1e1e

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/272/changes/f4395bee44d2dff21ffc09378f03dc4789713817

valkey_glide_cluster.stub.php
```diff
@@ -584,6 +584,11 @@ public function bgSave(mixed $route, ?string $mode = null): ValkeyGlideCluster|a
      */
     public function bgRewriteAof(mixed $route): ValkeyGlideCluster|array|bool|string;
 
+    /**
+     * @see ValkeyGlide::save
+     */
+    public function save(mixed $route): ValkeyGlideCluster|bool|string;
```
Aryex, line 590: is this missing an array type for multinode response

prateek-kumar-improving: SAVE always returns a scalar response (bool or string) even with multi-node routes. This is confirmed across Java, Python, Go, and C# implementations — none of them return a multi-node map for SAVE.

tests/ValkeyGlideClusterTest.php
currantw: Sounds good. Agreed that this will lead to a better error message in case of failure.

valkey_glide_cluster.stub.php
currantw: Some comments:

1. In all those other clients, a failure in _any_ node will produce an exception – so it's a bit different, no?
2. What is the behaviour here if the save fails on some nodes but succeeds on others?
3. Does this match the PHPRedis interface? What is it's behaviour?
4. Should we have separate documentation for the cluster clients that specifies what happens when called with a multi-node route?

currantw (commented):
Reviewed to f4395be.
Waiting on resolution on @Aryex 's comment (and any changes that may come from it) to approve.

Aryex (approved):

valkey_glide_cluster.stub.php
prateek-kumar-improving: About the PHPRedis part, it doesn't have RedisCluster::save() so there's no existing interface to match. So, the current behaviour is the same as other clients from glide-core side.

prateek-kumar-improving: > 1. In all those other clients, a failure in any node will produce an exception – so it's a bit different, no?

At the glide core level, if **any node fails** → an error is returned (which we convert to false). We just return false because of PHPRedis. So, it is a bit different to match PHPRedis for consistency.

> 2. What is the behaviour here if the save fails on some nodes but succeeds on others?

If any nodes fail, it returns an error. It is an error case. There are no partial success responses.

> 3. Does this match the PHPRedis interface? What is it's behaviour?

PHPRedis doesn't have RedisCluster::save()

> 4. Should we have separate documentation for the cluster clients that specifies what happens when called with a multi-node route?

We can but we will need to update it for all cluster commands. Maybe this part needs to be added in the official client documentation. This documentation can be common for all clients with some variations for PHP.

currantw: Thanks for the answers!

Hmmmm. I guess that is OK. Certainly that is the behaviour that I think most devs would guess. I wish there was a way to make this more explicit cleanly, but can't think of any now.

Please resolve on read.

currantw (approved):
Reviewed to f4395be.
Looks good ✅

PR #274, opened 2026-07-22
Implement client Pause and Unpause command

prateek-kumar-improving:
### Summary

Implement CLIENT PAUSE and CLIENT UNPAUSE commands for both standalone and cluster clients. CLIENT PAUSE suspends all client commands for a specified timeout, and CLIENT UNPAUSE resumes processing.

### Issue link

- Add ValkeyGlide::clientPause(int $timeout, ?string $mode = null): ValkeyGlide|bool|string
- Add ValkeyGlide::clientUnpause(): ValkeyGlide|bool|string
- Add ValkeyGlideCluster::clientPause(int $timeout, ?string $mode = null): ValkeyGlideCluster|bool|string
- Add ValkeyGlideCluster::clientUnpause(): ValkeyGlideCluster|bool|string
- With OPT_REPLY_LITERAL enabled, returns "OK" string
- Supports pause modes: "ALL" (default, blocks all commands) and "WRITE" (blocks only writes)
- In cluster mode, glide-core routes to all primary nodes by default (no explicit route param needed)

### Implementation

- execute_client_pause_command() in valkey_glide_commands.c — takes timeout (long) + optional mode (string)
- execute_client_unpause_command() in valkey_glide_commands.c — zero-arg command
- CLIENT_PAUSE_METHOD_IMPL and CLIENT_UNPAUSE_METHOD_IMPL macros in valkey_glide_commands_common.h
- ClientPause uses prepare_message_args (handles long + optional string args)
- ClientUnpause uses prepare_zero_args
- Both use execute_and_handle_batch helper and status processors

### Testing

Standalone tests (tests/ValkeyGlideTest.php) added.
Cluster tests (tests/ValkeyGlideClusterTest.php) added.

### Checklist

### DIFF
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -665,8 +665,8 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testClientPauseAndUnpause()
     {
-        // Pause then immediately unpause
-        $this->assertTrue($this->valkey_glide->clientPause(5000));
+        // Use a short pause so it expires before the unpause command times out
+        $this->assertTrue($this->valkey_glide->clientPause(200));
         $this->assertTrue($this->valkey_glide->clientUnpause());
     }
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -672,16 +672,18 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $result = $this->valkey_glide->clientPause($pauseMs, 'WRITE');
         $this->assertTrue($result);
 
-        // Reads should still work quickly during WRITE pause
-        $start = microtime(true);
-        $value = $this->valkey_glide->get($key);
-        $elapsed = (microtime(true) - $start) * 1000;
+        try {
+            // Reads should still work quickly during WRITE pause
+            $start = microtime(true);
+            $value = $this->valkey_glide->get($key);
+            $elapsed = (microtime(true) - $start) * 1000;
 
-        $this->assertEquals('value', $value);
-        $this->assertLT($pauseMs, $elapsed);
+            $this->assertEquals('value', $value);
+            $this->assertLT($pauseMs, $elapsed);
+        } finally {
+            $this->valkey_glide->clientUnpause();
+        }
 
-        // Unpause to restore normal operation
-        $this->valkey_glide->clientUnpause();
         $this->valkey_glide->del($key);
     }
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -659,6 +659,11 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testClientPauseWithMode()
     {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT PAUSE WRITE requires 6.2.0+');
+            return;
+        }
+
         $key = '{client_pause_write}_' . uniqid();
         $this->valkey_glide->set($key, 'value');
 
@@ -682,6 +687,11 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testClientUnpause()
     {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT PAUSE WRITE requires 6.2.0+');
+            return;
+        }
+
         $key = '{client_unpause}_' . uniqid();
         $this->valkey_glide->set($key, 'before');
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -675,17 +675,15 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testClientPauseWithReplyLiteral()
     {
-        $this->valkey_glide->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
+        $this->withOptReplyLiteralEnabled(function () {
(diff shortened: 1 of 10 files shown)


valkey_glide_cluster.stub.php
```diff
@@ -971,6 +971,16 @@ public function pfmerge(string $key, array $keys): ValkeyGlideCluster|bool;
      */
     public function ping(mixed $route, ?string $message = null): mixed;
 
+    /**
+     * @see ValkeyGlide::clientPause
+     */
+    public function clientPause(int $timeout, ?string $mode = null): ValkeyGlideCluster|bool|string;
```
Aryex, line 977: Should these commands take explicit routing in cluster mode? By default it is routed to random note .

prateek-kumar-improving: CLIENT PAUSE and CLIENT UNPAUSE have a built-in default route of AllPrimaries in glide-core. So, they're always sent to all primary nodes automatically. This matches the Java, Python, and C# implementations which also don't take a route parameter for these commands.

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,101 @@ public function testBgRewriteAofBatch()
...
     }
 
+    public function testClientPause()
+    {
+        $key = '{client_pause_all}_' . uniqid();
+        $this->valkey_glide->set($key, 'before');
+
+        $pauseMs = 2000;
+        $result = $this->valkey_glide->clientPause($pauseMs);
+        $this->assertTrue($result);
```
jamesx-improving, line 647: These cluster tests call `clientPause`/`clientUnpause` with no route and expect a scalar `true`/`"OK"`, but `process_core_cluster_result` returns a per-node array when the response is a `Map` — e.g. `testBgSave('allPrimaries')` asserts `assertIsArray($result)`. Could you confirm glide-core aggregates the default-route response for `ClientPause`/`ClientUnpause` into a single `Ok`? If not, these five tests need either an explicit route (e.g. `'randomNode'`) or per-node array assertions.

Affects `testClientPause`, `testClientPauseWithMode`, `testClientUnpause`, `testClientPauseWithReplyLiteral`, `testClientPauseBatch`.

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2721,101 @@ public function testBgRewriteAofBatch()
...
+    }
+
+    public function testClientPauseWithMode()
+    {
+        $key = 'client_pause_write_' . $this->createRandomString();
+        $this->valkey_glide->set($key, 'value');
+
+        // Pause with WRITE mode - only writes are blocked, reads still allowed
+        $pauseMs = 60000;
+        $result = $this->valkey_glide->clientPause($pauseMs, 'WRITE');
```
jamesx-improving, line 2756: `CLIENT PAUSE ... WRITE` was added in Redis/Valkey 6.2.0, but this test (and `testClientUnpause` below, plus the cluster equivalents) sends `'WRITE'` unconditionally. `testBgSaveCancel` uses `minVersionCheck('8.1.0')` for a similar version-gated feature — worth adding the same guard here if the CI matrix still touches pre-6.2 servers.

**Suggested fix:** wrap with `if (!$this->minVersionCheck('6.2.0')) { $this->markTestSkipped(...); return; }`.

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2721,101 @@ public function testBgRewriteAofBatch()
...
+        $this->valkey_glide->del($key);
+    }
+
+    public function testClientPauseWithMode()
+    {
+        $key = 'client_pause_write_' . $this->createRandomString();
+        $this->valkey_glide->set($key, 'value');
+
+        // Pause with WRITE mode - only writes are blocked, reads still allowed
+        $pauseMs = 60000;
```
jamesx-improving, line 2755: The 60-second `clientPause(60000, 'WRITE')` isn't wrapped in `try/finally`, so a real exception between here and `clientUnpause()` on line 2763 leaves the server-side pause active for up to a minute — every subsequent write-test would block. Setup/teardown reconnects the client, but the pause is server-side, not connection-scoped.

**Suggested fix:** wrap in `try { ... } finally { $this->valkey_glide->clientUnpause(); }`, or drop the pause to ~2000ms.

Same pattern in `tests/ValkeyGlideClusterTest.php:667`.

jamesx-improving (commented):
Nice, clean reuse of the existing unified-command framework — the C-level implementation mirrors `reset` / `bgSave` and I have no correctness or security concerns there. Left a few unverified test-side questions inline that I couldn't confirm without an initialized glide-core submodule; happy to re-review once you've had a look.

Aryex (approved):

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Checked on glide-core end, it returns OK from glide core side:
1. Glide core sends CLIENT PAUSE to all primaries
2. Waits for all to respond OK
3. Returns a single OK back to the PHP client (not a map of responses)

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/274/changes/4ba262899c9acbd00fcf8bb9dece889258d5d94e

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/274/changes/5e8718a30cd153f89ad16ae6bb8eb3ad4a6d4ec5

jamesx-improving (approved):
LGTM, thanks for the fixes.

Issue #275, opened 2026-07-22
[Task] Add a second standalone Valkey server instance to the CI test infrastructure


prateek-kumar-improving:
### Summary
Add a second standalone Valkey server instance to the CI test infrastructure to enable full integration testing of the MIGRATE command with actual key transfers.

### Context

The MIGRATE command (#275) is implemented and tested for error paths (invalid host, NOKEY for non-existent keys, COPY/REPLACE options). However, testing successful key migration requires a second Valkey 
instance to migrate to, which our CI currently doesn't provide.

Python and C# GLIDE clients have this infrastructure (Python uses a second_server fixture, C# uses TestClients), enabling tests that verify:
- Key is transferred to the destination
- Key is deleted from the source (unless COPY is set)
- Multi-key migration works end-to-end

### Test Scenarios to Add

Tests requiring a second Valkey server instance for migrate command:

1. Single-key migrate success (standalone) — migrate key, verify deleted from source, verify exists on destination
2. Single-key migrate success (cluster) — same as above but on cluster client
3. Multi-key migrate success (standalone) — migrate multiple keys, verify all moved
4. Migrate with COPY flag success — migrate key, verify it still exists on source AND exists on destination
5. Migrate with REPLACE flag success — set a key on destination first, migrate from source, verify destination value is overwritten
6. Migrate with AUTH success — destination requires password, migrate with credentials
7. Migrate with AUTH2 success — destination requires username+password, migrate with AUTH2 credentials

php
// Single-key successful migration
$this->valkey_glide->set('migrate_key', 'value');
$result = $this->valkey_glide->migrate('localhost', 6389, 'migrate_key', 0, 5000);
$this->assertTrue($result);
$this->assertFalse($this->valkey_glide->exists('migrate_key'));
// Verify on destination: key exists with correct value

// With COPY flag
$result = $this->valkey_glide->migrate('localhost', 6389, 'key', 0, 5000, true);
$this->assertTrue($result);
$this->assertTrue($this->valkey_glide->exists('key')); // Still exists on source

### Related

- MIGRATE implementation: #275
- Python reference: test_sync_migrate_success in test_sync_client.py
- C# reference: GenericCommandTests in PR #447

prateek-kumar-improving:
Fixing this issue in this PR: https://github.com/valkey-io/valkey-glide-php/pull/276 and updating the tests with the same PR as well.

PR #276, opened 2026-07-22
Implement migrate command

prateek-kumar-improving:
### Summary
Implement the MIGRATE command for both standalone and cluster clients. Supports single-key migration (both clients), multi-key migration (standalone only), COPY/REPLACE flags, and AUTH/AUTH2 credentials.

### Issue link

Linked issue: https://github.com/valkey-io/valkey-glide-php/issues/275

* Add ValkeyGlide::migrate(string $host, int $port, string|array $key, int $dstdb, int $timeout, bool $copy = false, bool $replace = false, mixed $credentials = null): ValkeyGlide|bool|string
* Add ValkeyGlideCluster::migrate(...): same but single-key only
* PHPRedis compatible signature
* Returns true on success (OK/NOKEY), false on error
* With OPT_REPLY_LITERAL enabled, returns raw status string ("OK" or "NOKEY") on success, false on failure
* Cluster mode: single-key only (matching Java/Python/Go limitation)
* Supports COPY, REPLACE, AUTH, and AUTH2

### Implementation
* execute_migrate_command() in valkey_glide_commands.c
* MIGRATE_METHOD_IMPL macro using STANDARD_METHOD_IMPL
* Migrate case added to prepare_message_args switch (alphabetized)
* Increased core_command_args_t.args from [8] to [12]
* Multi-key handled by iterating PHP array as individual STRING args
* Credentials support: string → AUTH, array [username, password] → AUTH2
* Test infrastructure: startDestinationServer()/stopDestinationServer() helpers using cluster_manager.py

### Testing

Standalone tests (tests/ValkeyGlideTest.php) added.
Cluster tests (tests/ValkeyGlideClusterTest.php) added.

### Checklist


valkey_glide.stub.php
```diff
@@ -2225,6 +2225,30 @@ public function mget(array $keys): ValkeyGlide|array|false;
...
+    /**
+     * Atomically transfer a key from the connected instance to a destination instance.
+     *
+     * On success, the key is deleted from the source instance (unless COPY is set).
+     * The command is atomic and blocks both instances during the transfer.
+     *
+     * @param string       $host        The destination host.
+     * @param int          $port        The destination port.
+     * @param string|array $key         The key to migrate, or an array of keys for multi-key migration.
+     * @param int          $dstdb       The destination database index.
```
currantw, line 2237: The other clients seem to use `$destinationDb` or equivalent. Why the choice to go with this shorter name here – is "dst" used elsewhere in the PHP client as an abbreviation for "destination"?

valkey_glide_cluster.stub.php
```diff
@@ -486,6 +486,11 @@ public function select(int $db): ValkeyGlideCluster|bool;
      */
     public function move(string $key, int $index): ValkeyGlideCluster|bool;
 
+    /**
+     * @see ValkeyGlide::migrate
+     */
```
currantw, line 493: Do we need to have difference documentation here to indicate the only single-key is supported?

valkey_glide_commands.c
```diff
@@ -359,6 +359,255 @@ int execute_bgrewriteaof_command(zval* object, int argc, zval* return_value, zen
     return execute_and_handle_batch(valkey_glide, &core_args, processor, return_value, object);
 }
 
+/* Execute a MIGRATE command using the Valkey Glide client - UNIFIED IMPLEMENTATION */
```
currantw: Aside: I've noticed the "UNIFIED IMPLEMENTATION" comment here on many commands. Is that necessary? Are there other "non-unified" implementations? What value is it adding? 🤔

valkey_glide_core_common.h
```diff
@@ -105,11 +105,13 @@ typedef struct {
     zval*            raw_options; /* Raw PHP options array for complex parsing */
     size_t           key_len;
     core_options_t   options;
-    core_arg_t       args[8]; /* Fixed arguments array - sufficient for current usage */
+    core_arg_t       args[12]; /* Fixed arguments array - supports most commands */
+    core_arg_t*      dyn_args; /* Dynamic overflow args (heap-allocated, for MIGRATE multi-key) */
```
currantw: Why did we increase the fixed argument size to 12 _**and**_ also add this dynamic overflow args?

tests/ValkeyGlideTest.php
```diff
@@ -88,6 +88,28 @@ private function createRandomString($length = 10)
         return bin2hex(random_bytes(ceil($length / 2)));
     }
 
+    /**
+     * Port for the destination Valkey server used in migrate tests.
+     */
+    protected const MIGRATE_DEST_PORT = 6382;
```
currantw, line 95: Nit. Maybe a reference to `tests/start_valkey_with_replicas.sh` would help?


```suggestion
    /**
     * Port for the destination Valkey server used in migrate tests.
     * See also "tests/start_valkey_with_replicas.sh".
     */
    protected const MIGRATE_DEST_PORT = 6382;
```

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2743,229 @@ public function testBgRewriteAofBatch()
         $this->assertTrue($result[0]);
     }
 
+    public function testMigrate()
```
currantw: This method name seems potentially misleading?

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2743,229 @@ public function testBgRewriteAofBatch()
...
+        $this->assertEquals('value1', $dest->get($key1));
+        $this->assertEquals('value2', $dest->get($key2));
+    }
+
+    public function testMigrateWithAuthCredentials()
+    {
+        // AUTH with string password - should not error (server may not require auth,
+        // but verifies the argument is correctly serialized and sent)
+        $key = 'migrate_auth_' . $this->createRandomString();
+        $this->valkey_glide->set($key, 'auth_value');
```
currantw: Does this actually "verifi[y that] the argument is correctly serialized and sent"? Do we have any existing testing for username/password authentication? If not, that seems like something we should add?

Perhaps this could be a separate issue? I imagine we would want to add a server with password and one with password + username, and to add additional tests for connecting to them?

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2743,229 @@ public function testBgRewriteAofBatch()
...
+
+        $result = $this->valkey_glide->migrate('127.0.0.1', self::MIGRATE_DEST_PORT, $key, 0, 5000);
+        $this->assertTrue($result);
+
+        // Key should be removed from source and exist at destination
+        $this->assertEquals(0, $this->valkey_glide->exists($key));
+        $this->assertEquals('migrate_value', $dest->get($key));
+    }
+
+    public function testMigrateMultiKeySuccess()
```
currantw, line 2827: Does this multi-key variant exercise the dynamic args logic?

valkey_glide_core_common.c
```diff
@@ -705,6 +718,44 @@ int prepare_message_args(core_command_args_t* args,
         }
     }
 
+    /* Add dynamic overflow arguments (used by MIGRATE multi-key) */
+    for (int i = 0; i < args->dyn_arg_count; i++) {
+        switch (args->dyn_args[i].type) {
```
currantw: I'm still a bit confused we we need a separate dynamic args section? Why can't the entire args array just be dynamically sized? Is this a performance optimization of some sort? Why does this only apply to this one command – do we not have any other multi-key commands?

currantw (commented):
Reviewed to 2c1ce7d.

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/b71b8e933721a09febb6f448e4c50ab64769af58

valkey_glide.stub.php
prateek-kumar-improving: It is the same in PHPRedis. But we can change it, as the name can be different. Updated in these commits: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/b71b8e933721a09febb6f448e4c50ab64769af58, https://github.com/valkey-io/valkey-glide-php/pull/276/changes/c9aa25a30a91ee1ea6d591dcc61997e7ed9fd091, https://github.com/valkey-io/valkey-glide-php/pull/276/changes/fbb982bb683400dc4262adf52336a914ba3977de

valkey_glide_commands.c
prateek-kumar-improving: This is something that was added in all previous comments for all these methods in the file. The code had 2 methods for some cases but it was cleanup at some point but now there are no "non-unified" implementations.

 Removing all of them as it does not add any value: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/fbb982bb683400dc4262adf52336a914ba3977de

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/fbb982bb683400dc4262adf52336a914ba3977de

tests/ValkeyGlideTest.php
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/572783fad09d41df10b15778aff376b70a3e9e7c

prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/572783fad09d41df10b15778aff376b70a3e9e7c

valkey_glide_core_common.h
prateek-kumar-improving: The fixed array was increased from 8 to 12 to accommodate MIGRATE's single-key case without heap allocation: `host, port, "", db, timeout, COPY, REPLACE, AUTH, password, KEYS`, key = 11 args max for single-key with auth.
The dyn_args was added separately for multi-key MIGRATE where the total arg count is unbounded (user can pass any number of keys). For single-key MIGRATE and all other commands, everything fits in the fixed array and dyn_args is unused.

prateek-kumar-improving: Removed dyn_args/dyn_arg_count. Replaced with single all_args pointer. When multi-key MIGRATE needs more than 12 args, it allocates one combined array. Single loop in prepare_message_args. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/991f1f10d790aff22ed516f1ced9684732e898d6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Fixed: Added testMigrateMultiKeyManyKeys — 10 keys + COPY + REPLACE = 19 args total, guaranteeing dynamic allocation is exercised.
Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/991f1f10d790aff22ed516f1ced9684732e898d6

valkey_glide_core_common.c
prateek-kumar-improving: 1. Simplified in the latest commit. Removed` dyn_args/dyn_arg_count` entirely. Now there's a single all_args pointer on the struct — if non-NULL, prepare_message_args iterates it 
instead of the fixed args[12].
2. For multi-key MIGRATE, execute_migrate_command allocates one combined `core_arg_t*` array with all arguments in order (fixed positional args + KEYS + keys), sets all_args to it.



3. DEL/MGET/EXISTS — the command is just: `DEL key1 key2 key3`
The entire argument list is only keys. So prepare_multi_key_args just takes the PHP array, allocates a cmd_args array of the right size, and copies the keys in. Simple — no mixing with other args.

MIGRATE — the command is: `MIGRATE host port "" db timeout COPY REPLACE AUTH password KEYS key1 key2 key3`

It has positional args first (host, port, empty string, db, timeout), then optional flags (COPY, REPLACE, AUTH), then the KEYS keyword, then the actual keys. The keys are embedded in the middle of a complex argument structure. That's why MIGRATE can't just use prepare_multi_key_args — it needs to build a single ordered array containing both the fixed positional/flag args AND the variable key list together.

valkey_glide_commands.c
```diff
@@ -359,7 +359,230 @@ int execute_bgrewriteaof_command(zval* object, int argc, zval* return_value, zen
...
+        }
+        /* Pass empty string as key placeholder, keys will be appended via KEYS keyword */
+        core_args.args[arg_idx].type                  = CORE_ARG_TYPE_STRING;
+        core_args.args[arg_idx].data.string_arg.value = "";
+        core_args.args[arg_idx].data.string_arg.len   = 0;
+    } else {
+        return 0;
+    }
+    arg_idx++;
+
```
Aryex, line 439: **Bug (blocking):** Single-element credentials array `['password']` is silently dropped.

The `ZEND_HASH_FOREACH_VAL` loop assigns element at `idx == 0` to `username`. For a single-element array, the loop ends with `password` still NULL. The subsequent `if (password && arg_idx + 2 < 12)` check fails, so AUTH is never appended. The "AUTH password (array with single element)" else-branch on line ~168 is dead code for genuine single-element arrays.

**Impact:** `migrate(..., credentials: ['pass'])` sends MIGRATE without AUTH — the destination server may reject with NOAUTH or, worse, allow an unauthenticated transfer.

**Suggested fix:** After the loop, check element count. If `idx == 1` (single element), treat the captured value as `password` (not `username`):

```c
ZEND_HASH_FOREACH_END();

// Single-element array: treat as AUTH password
if (idx == 1 && username && !password) {
    password = username;
    password_len = username_len;
    username = NULL;
    username_len = 0;
}
```

Alternatively, use `zend_hash_index_find(ht, 0)` / `zend_hash_index_find(ht, 1)` and branch on `zend_hash_num_elements(ht)` for clarity.

valkey_glide_commands.c
```diff
@@ -359,7 +359,230 @@ int execute_bgrewriteaof_command(zval* object, int argc, zval* return_value, zen
...
+    if (z_credentials && Z_TYPE_P(z_credentials) != IS_NULL) {
+        if (Z_TYPE_P(z_credentials) == IS_STRING) {
+            /* Simple password: AUTH password (need 2 slots) */
+            if (arg_idx + 1 < 12) {
+                core_args.args[arg_idx].type                  = CORE_ARG_TYPE_STRING;
+                core_args.args[arg_idx].data.string_arg.value = "AUTH";
+                core_args.args[arg_idx].data.string_arg.len   = 4;
+                arg_idx++;
+                core_args.args[arg_idx].type                  = CORE_ARG_TYPE_STRING;
+                core_args.args[arg_idx].data.string_arg.value = Z_STRVAL_P(z_credentials);
```
Aryex, line 476: **Bug (blocking):** Non-string elements in the keys array are silently skipped.

This `if (Z_TYPE_P(z_val) == IS_STRING)` check means integer keys (which are valid Valkey keys and common in PHP — e.g., `[123, 'user:1']`) are quietly dropped. The command still returns `true` even though not all keys were migrated. This is a silent data-integrity issue.

PHPRedis coerces non-string values to strings before sending them.

**Suggested fix:** Coerce with `zval_get_string()` (or the non-allocating `zval_get_tmp_string()`):

```c
zval* z_val;
ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(z_key), z_val) {
    zend_string *tmp_str, *str = zval_get_tmp_string(z_val, &tmp_str);
    all[arg_idx].type                  = CORE_ARG_TYPE_STRING;
    all[arg_idx].data.string_arg.value = ZSTR_VAL(str);
    all[arg_idx].data.string_arg.len   = ZSTR_LEN(str);
    arg_idx++;
    // Note: tmp_str lifetime is valid until zend_tmp_string_release()
    // You may need to track these for release after execute_and_handle_batch
}
ZEND_HASH_FOREACH_END();
```

If lifetime management is complex, an alternative is to reject non-string elements with `zend_argument_value_error()` instead of silently skipping them.

tests/start_valkey_with_replicas.sh
```diff
@@ -37,6 +37,15 @@ valkey-server --port 6381 \
   --logfile "$BASE_DIR/6381/valkey.log" \
   --enable-debug-command yes
 
+# Start independent standalone server for migrate tests (6382)
+valkey-server --port 6382 \
```
Aryex, line 41: **Infra bug (blocking):** Port 6382 is started here but never stopped in `stop_valkey_replicas.sh` (the shutdown loop only covers 6379-6381).

This leaves an orphaned `valkey-server` process bound to `0.0.0.0:6382` with `protected-mode no` after every test run. On CI runners this accumulates; locally it causes the next `start` to silently reuse a stale server.

**Fix:** Add 6382 to the `for port in ...` loop in the stop script.

Aryex (changes requested):

tests/ValkeyGlideTest.php
prateek-kumar-improving: Currently, it just tests that it does not throw an error. There are no such tests across all languages.
Not sure if this is something any client wants to test, so, not sure if we want to create an issue for something that will never be implemented.

valkey_glide_commands.c
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/3ffa4e0d7fc32d1c3d799cbee41a42dc87d8edc6

valkey_glide_commands.c
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/3ffa4e0d7fc32d1c3d799cbee41a42dc87d8edc6

tests/start_valkey_with_replicas.sh
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/3ffa4e0d7fc32d1c3d799cbee41a42dc87d8edc6

valkey_glide.stub.php
currantw: > as the name can be different

Is that true? It seems to me that PHP 8+ supports named arguments. If that's true, I think we need to match PHPRedis? 🤔

tests/ValkeyGlideTest.php
currantw: Thanks!

tests/ValkeyGlideTest.php
currantw: I think this is a significant gap. I have raised [an issue](https://github.com/valkey-io/valkey-glide-csharp/issues/487) to add these for C# and will be prioritizing for for Release 1.2, and I think this is something that we should certainly have for all clients.

valkey_glide_commands.c
currantw: This illustrates exactly why I think it is important to have test coverage for password and username/password authentication – these tests could have caught this (and they will catch any regressions in the future).

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,148 @@ public function testBgRewriteAofBatch()
...
+    public function testMigrateToInvalidHostReturnsFalse()
+    {
+        $key = '{migrate_test}_' . uniqid();
+        $this->valkey_glide->set($key, 'test_value');
+
+        // Existing key + invalid host returns false (connection error)
+        $result = $this->valkey_glide->migrate('nonexistent.invalid', 9999, $key, 0, 1000);
+        $this->assertFalse($result);
+
+        $this->valkey_glide->del($key);
```
currantw: Is this common in other tests? One of the advantages of using a random key is, I think, that you don't actually need to delete it after? 🤷

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,148 @@ public function testBgRewriteAofBatch()
...
+    }
+
+    public function testMigrateNonExistentKeyReturnsTrue()
```
currantw: This it makes sense to to use the valid migration database for this test?

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,148 @@ public function testBgRewriteAofBatch()
...
+            'nonexistent.invalid',
+            6379,
+            'nonexistent_key_' . uniqid(),
+            0,
+            1000
+        );
+        $this->assertTrue($result);
+    }
+
+    public function testMigrateNonExistentKeyWithReplyLiteral()
```
currantw: Same comment as above. Think it makes sense to use the valid migration DB here to make sure we are testing the expected case.

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,148 @@ public function testBgRewriteAofBatch()
...
+        $dest = $this->getMigrateDestClient();
+
+        $key = '{migrate_copy}_' . uniqid();
+        $this->valkey_glide->set($key, 'copy_value');
+
+        $result = $this->valkey_glide->migrate('127.0.0.1', self::MIGRATE_DEST_PORT, $key, 0, 5000, true);
+        $this->assertTrue($result);
+
+        // Key should remain at source and also exist at destination
+        $this->assertEquals(1, $this->valkey_glide->exists($key));
```
currantw: Nit. Don't think this `exists` call is necessary – won't `get` already fail if the key doesn't exist?

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,148 @@ public function testBgRewriteAofBatch()
...
+        // Set a new value at source
+        $this->valkey_glide->set($key, 'new_source_value');
+
+        // Migrate with REPLACE - should overwrite destination
+        $result = $this->valkey_glide->migrate('127.0.0.1', self::MIGRATE_DEST_PORT, $key, 0, 5000, false, true);
+        $this->assertTrue($result);
+
+        // Key should be removed from source, destination has new value
+        $this->assertEquals(0, $this->valkey_glide->exists($key));
+        $this->assertEquals('new_source_value', $dest->get($key));
```
currantw: Nit. This seems more complex than it needs to be. Why can't we just set the key that will be overwritten directly on the migrate database (rather than setting it on the source DB and migrating it)?

tests/ValkeyGlideClusterTest.php
```diff
@@ -637,6 +637,148 @@ public function testBgRewriteAofBatch()
...
+
+    public function testMigrateBatchMultiKey()
+    {
+        $this->markTestSkipped('Multi-key migration not supported in cluster mode');
+    }
+
+    public function testMigrateEmptyKeys()
+    {
+        $this->markTestSkipped('Multi-key migration not supported in cluster mode');
+    }
```
currantw, line 668: These can be removed. Can you please make sure to verify changes made by AI agents?

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2744,265 @@ public function testBgRewriteAofBatch()
...
+
+        // Create 10 keys to exercise dynamic allocation (total args > 12)
+        $keys = [];
+        for ($i = 0; $i < 10; $i++) {
+            $key = 'migrate_many_' . $i . '_' . $this->createRandomString();
+            $this->valkey_glide->set($key, "value_$i");
+            $keys[] = $key;
+        }
+
+        // Migrate with COPY + REPLACE — args: host, port, "", db, timeout, COPY, REPLACE, KEYS, key0..key9
```
currantw: Nit. Why do we need to include all the args in this comment?

tests/ValkeyGlideTest.php
```diff
@@ -2721,6 +2744,265 @@ public function testBgRewriteAofBatch()
...
+        $this->valkey_glide->set($key, 'auth2_value');
+
+        // Migrate with AUTH2 credentials to invalid host - verifies credentials are passed correctly
+        $result = $this->valkey_glide->migrate('nonexistent.invalid', 9999, $key, 0, 1000, false, false, ['myuser', 'mypassword']);
+        $this->assertFalse($result);
+
+        $this->valkey_glide->del($key);
+    }
+
+    public function testMigrateWithAuth2SingleElementArray()
```
currantw, line 2992: We do not appear to have AUTH tests for cluster?

tests/ValkeyGlideTest.php
currantw: If we are going to keep these tests as is, I think it is important that the comments are clear about what they do and do not test. As written, then verify that these AUTH arguments do not throw errors. They do not verify that the argument is correctly serialized and sent, nor that "credentials are passed correctly".

valkey_glide.stub.php
```diff
@@ -2225,6 +2225,39 @@ public function mget(array $keys): ValkeyGlide|array|false;
      */
     public function move(string $key, int $index): ValkeyGlide|bool;
 
+    /**
+     * Atomically transfer a key from the connected instance to a destination instance.
```
currantw: Nit.


```suggestion
     * Atomically transfers one or more key from the connected instance to a destination instance.
```

valkey_glide.stub.php
```diff
@@ -2225,6 +2225,39 @@ public function mget(array $keys): ValkeyGlide|array|false;
...
+     *
+     * On success, the key is deleted from the source instance (unless COPY is set).
+     * The command is atomic and blocks both instances during the transfer.
```
currantw:
```suggestion
     * On success, the keys are deleted from the source instance (unless COPY is set).
     * The command is atomic and blocks both instances during the transfer.
```

valkey_glide.stub.php
```diff
@@ -2225,6 +2225,39 @@ public function mget(array $keys): ValkeyGlide|array|false;
...
+     * @param bool         $copy        If true, do not remove the key from the source instance.
+     * @param bool         $replace     If true, replace existing key on the destination instance.
+     * @param mixed        $credentials Optional authentication: a string password, or
+     *                                  an array [username, password] for AUTH2.
+     *
+     * @return ValkeyGlide|bool|string  Returns true on success, false on failure.
+     *                                  With OPT_REPLY_LITERAL enabled, returns "OK" or "NOKEY" on success,
+     *                                  false on failure.
+     *
+     * @see https://valkey.io/commands/migrate
```
currantw, line 2248:
```suggestion
     * @param string       $host        The destination host.
     * @param int          $port        The destination port.
     * @param string|array $key         The key to migrate, or an array of keys for multi-key migration.
     * @param int          $destinationDb       The destination database index.
     * @param int          $timeout     The maximum idle time in milliseconds for the transfer.
     * @param bool         $copy        If true, do not remove the key(s) from the source instance.
     * @param bool         $replace     If true, replace existing key(s) on the destination instance.
     * @param mixed        $credentials Optional authentication: a string password, or
     *                                  an array [username, password] for AUTH2.
     *
     * @return ValkeyGlide|bool|string  Returns true on success, false on failure.
     *                                  With OPT_REPLY_LITERAL enabled, returns "OK" or "NOKEY" on success,
     *                                  false on failure.
     *
     * @see https://valkey.io/commands/migrate
```

valkey_glide_core_common.c
currantw: Hmmm. I find this entire quagmire deeply unsatisfying. I seems like the entire structure of valkey_glide_core_common is very convoluted and probably needs a re-think. That being said, I don't have any better suggestions for now. Please resolve.

currantw (commented):
Reviewed to 6296505.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: They can't be removed. If removed, these tests will run for both standalone and cluster. That's how it is in the language.

tests/ValkeyGlideTest.php
prateek-kumar-improving: Created an issue for PHP: https://github.com/valkey-io/valkey-glide-php/issues/286

valkey_glide.stub.php
prateek-kumar-improving: Updated back to the original name here: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/379925f65a87ba8c6b89c04709d8067267e9d712

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/a46eb184d4da321ab2e57b36f9b0cc572eab6a26

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: I think it would not matter for the testing but it would make it close to the real scenario. Updating in these commits: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/ea95bafca7e3b379f75f35a2f778e9e71e455760, https://github.com/valkey-io/valkey-glide-php/pull/276/changes/683b12e04c98a82941b6c4ac75cb1119e67e1f8b

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in these commits: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/ea95bafca7e3b379f75f35a2f778e9e71e455760, https://github.com/valkey-io/valkey-glide-php/pull/276/changes/683b12e04c98a82941b6c4ac75cb1119e67e1f8b

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/66cbea71a5748c5e04ee16e4ad2b17123ec4d953

tests/ValkeyGlideTest.php
prateek-kumar-improving: Added comments in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/4ac362afe1cb128ef903442ecc8fb1983ae73640

valkey_glide.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/bc9a734245d3078cd2b0a79798b4a6518656ba1e

valkey_glide.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/bc9a734245d3078cd2b0a79798b4a6518656ba1e

valkey_glide.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/bc9a734245d3078cd2b0a79798b4a6518656ba1e

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/441a10e3385e5e77d7da5fd54bea45625a86b4ea

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/100f017df548f9565246470d39070e9b196ea541

tests/ValkeyGlideTest.php
prateek-kumar-improving: The auth tests are in ValkeyGlideTest.php and are inherited by ValkeyGlideClusterTest (which extends it), so they run in both standalone and cluster mode automatically.

Aryex (approved):
lgtm

tests/ValkeyGlideClusterTest.php
currantw: > I think it would not matter for the testing but it would make it close to the real scenario

Agreed that it won't matter. _We_ know that it won't matter for the test because _we_ know the internal implementation details. But I think that the test should generally (if possible) avoid assumption that are based on this information. Please feel free to resolve once read.

tests/ValkeyGlideClusterTest.php
currantw: Oh WOW, my bad!

Are there any plans to create a base test client class to avoid this? Seems like it would introduce a lot of unnecessary boilerplate? Is this something you can raise?

Feel free to resolve?

tests/ValkeyGlideTest.php
currantw: Gotcha, thanks

currantw: Wait a second. If `ValkeyGlideClusterTest` extends `ValkeyGlideTest` and inherits all its tests, why do the following test methods need to be defined for cluster?

- testMigrateBatch
- testMigrateWithReplaceOverwritesDestination
- testMigrateWithCopyKeyRemainsAtSource
- testMigrateSingleKeySuccess
- testMigrateNonExistentKeyWithReplyLiteral
- testMigrateNonExistentKeyReturnsTrue
- testMigrateToInvalidHostReturnsFalse

Shouldn't these test cases already be defined for standalone?

currantw (commented):
Reviewed to 100f017.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: No plans for now. Creating an issue: https://github.com/valkey-io/valkey-glide-php/issues/287. Will check later based on bandwidth if it is possible to remove these empty tests.

currantw (commented):
Just one remaining comment about possible duplicate tests

tests/ValkeyGlideTest.php
prateek-kumar-improving: They have different key names.
For single-key MIGRATE, key names with hash tags aren't necessary since there's no multi-slot concern. These cluster overrides duplicate the standalone tests with the only difference being {hash_tag} key prefixes. 

But since it is restricted to only single key operation in cluster, we don't have any slot concern for now in cluster. We can remove these tests for now.

prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/276/changes/8ce99a9bec22f53b0dddb472d92828b2b07d9592

currantw (approved):
Reviewed to 8ce99a9.
Looks good ✅ 
Thanks @prateek-kumar-improving 🎉

PR #277, opened 2026-07-23
Add circuit breaker configuration support

### LINKED ISSUE #238: [Task] Add circuit breaker configuration support for PHP
### Description

Add `ClientCircuitBreakerConfiguration` to the client configuration, matching the existing implementation in Java, Python, Node, and Go (valkey-io/valkey-glide#6050). The circuit breaker logic lives in the Rust core, this task only requires adding the config class, serializing it to protobuf, and surfacing a `CircuitBreakerException` type to callers.

### Checklist

- [ ] Add `ClientCircuitBreakerConfiguration` class
- [ ] Serialize config to protobuf `ConnectionRequest`
- [ ] Add `CircuitBreakerException` type
- [ ] Add unit test for config serialization

### Related

- valkey-io/valkey-glide#6050
- valkey-io/valkey-glide#5996

jeremyprime:
### Summary

Add circuit breaker configuration support to the PHP client, matching the existing implementation in Java, Python, Node, and Go (valkey-io/valkey-glide#6050). The circuit breaker logic lives in the Rust core; this change adds the config array parameter, serializes it to protobuf, and surfaces a `CircuitBreakerException` type to callers.

### Issue link

Closes #238

- New `?array $circuit_breaker` parameter on `ValkeyGlide::connect()` and `ValkeyGlideCluster::__construct()`
- New `CircuitBreakerException` class (extends `ValkeyGlideException`) thrown when the circuit breaker is open
- Configuration keys: `window_size_ms`, `failure_rate_threshold`, `min_errors`, `open_timeout_ms`, `count_timeouts`, `consecutive_successes`
- All parameters optional with sensible defaults matching other clients

### Implementation

- `common.h`: Added `valkey_glide_circuit_breaker_config_t` struct, constants, and plumbed into base config
- `valkey_glide.c`: Parse circuit breaker array, register exception class, cleanup on free
- `valkey_glide_core_commands.c`: Serialize config to `ClientCircuitBreakerConfig` protobuf message (field 30)
- `valkey_glide_commands_common.h`: Dispatch `CircuitBreakerException` when `command_error_type == CircuitBreakerOpen`
- Stubs and mock constructors updated for both standalone and cluster

### Testing

- New unit tests added to `ConnectionRequestTest.php`:
  - `testCircuitBreakerStandaloneDefaults`: empty array uses defaults
  - `testCircuitBreakerClusterDefaults`: same for cluster
  - `testCircuitBreakerStandaloneCustomValues`: all fields customized
  - `testCircuitBreakerClusterCustomValues`: same for cluster
  - `testCircuitBreakerNotSetByDefault`: null when not provided

### Checklist

### DIFF
--- a/common.h
+++ b/common.h
@@ -145,6 +145,31 @@ typedef struct {
     bool has_max_decompressed_size; /* true if user explicitly set max_decompressed_size */
 } valkey_glide_compression_config_t;
 
+/* Circuit breaker configuration */
+#define VALKEY_GLIDE_CB_WINDOW_SIZE_MS "window_size_ms"
+#define VALKEY_GLIDE_CB_FAILURE_RATE_THRESHOLD "failure_rate_threshold"
+#define VALKEY_GLIDE_CB_MIN_ERRORS "min_errors"
+#define VALKEY_GLIDE_CB_OPEN_TIMEOUT_MS "open_timeout_ms"
+#define VALKEY_GLIDE_CB_COUNT_TIMEOUTS "count_timeouts"
+#define VALKEY_GLIDE_CB_CONSECUTIVE_SUCCESSES "consecutive_successes"
+
+#define VALKEY_GLIDE_CB_DEFAULT_WINDOW_SIZE_MS 10000
+#define VALKEY_GLIDE_CB_DEFAULT_FAILURE_RATE_THRESHOLD 0.5
+#define VALKEY_GLIDE_CB_DEFAULT_MIN_ERRORS 50
+#define VALKEY_GLIDE_CB_DEFAULT_OPEN_TIMEOUT_MS 5000
+#define VALKEY_GLIDE_CB_DEFAULT_COUNT_TIMEOUTS false
+#define VALKEY_GLIDE_CB_DEFAULT_CONSECUTIVE_SUCCESSES 3
+#define VALKEY_GLIDE_ERROR_TYPE_CIRCUIT_BREAKER_OPEN 4
+
+typedef struct {
+    uint32_t window_size_ms;
+    float    failure_rate_threshold;
+    uint32_t min_errors;
+    uint32_t open_timeout_ms;
+    bool     count_timeouts;
+    uint32_t consecutive_successes;
+} valkey_glide_circuit_breaker_config_t;
+
 /* Client-side cache configuration */
 typedef struct {
     char*  cache_id;     /* Unique cache identifier */
@@ -187,6 +212,7 @@ typedef struct {
     valkey_glide_advanced_base_client_configuration_t* advanced_config;    /* NULL if not set */
     valkey_glide_compression_config_t*                 compression_config; /* NULL if not set */
     valkey_glide_client_side_cache_config_t*           client_side_cache;  /* NULL if not set */
+    valkey_glide_circuit_breaker_config_t*             circuit_breaker;    /* NULL if not set */
     valkey_glide_read_from_t                           read_from;
     int                                                addresses_count;
     int                                                request_timeout;         /* -1 if not set */
@@ -222,6 +248,7 @@ typedef struct {
     zval*     context;           /* Stream context for TLS */
     zval*     compression;       /* Compression configuration */
     zval*     client_side_cache; /* Client-side cache configuration */
+    zval*     circuit_breaker;   /* Circuit breaker configuration */
     char*     client_name;
     char*     client_az;
     size_t    client_name_len;
@@ -318,6 +345,7 @@ PHP_MINFO_FUNCTION(redis);
 
 zend_class_entry* get_valkey_glide_ce(void);
 zend_class_entry* get_valkey_glide_exception_ce(void);
+zend_class_entry* get_valkey_glide_circuit_breaker_exception_ce(void);
 
 zend_class_entry* get_valkey_glide_cluster_ce(void);
 
--- a/tests/ConnectionRequestTest.php
+++ b/tests/ConnectionRequestTest.php
@@ -75,6 +75,7 @@ defined('VALKEY_GLIDE_PHP_TESTRUN') or die("Use TestValkeyGlide.php to run tests
 require_once __DIR__ . "/TestSuite.php";
 require_once __DIR__ . "/../vendor/autoload.php";
 require_once __DIR__ . "/Connection_request/AuthenticationInfo.php";
+require_once __DIR__ . "/Connection_request/ClientCircuitBreakerConfig.php";
 require_once __DIR__ . "/Connection_request/CompressionBackend.php";
 require_once __DIR__ . "/Connection_request/CompressionConfig.php";
 require_once __DIR__ . "/Connection_request/ConnectionRequest.php";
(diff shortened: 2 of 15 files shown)


valkey_glide_commands_common.h
```diff
@@ -74,8 +82,8 @@ static inline void handle_command_result_or_throw(CommandResult* result,
         return;
     }
     if (result->command_error) {
-        zend_throw_exception(
-            get_valkey_glide_exception_ce(), result->command_error->command_error_message, 0);
+        zend_class_entry* exception_ce = get_exception_ce_for_command_error(result->command_error);
+        zend_throw_exception(exception_ce, result->command_error->command_error_message, 0);
```
prateek-kumar-improving: Should all zend_throw_exception method calls in the codebase be updated to use `get_exception_ce_for_command_error()`? Can the `command_error_type` be returned at other zend_throw_exception() calls that are not updated?

tests/ConnectionRequestTest.php
```diff
@@ -742,4 +743,94 @@ private function get_addresses(\Connection_request\ConnectionRequest $request):
     {
         return array_map(fn($a) => $a->getHost(), iterator_to_array($request->getAddresses()));
     }
+
+    // ================================================================
+    // Circuit Breaker Tests
+    // ================================================================
```
prateek-kumar-improving, line 749: Please add edge-case tests for invalid values.

valkey_glide.c
```diff
@@ -573,6 +579,60 @@ int valkey_glide_build_client_config_base(valkey_glide_php_common_constructor_pa
         config->client_side_cache = NULL;
     }
```
prateek-kumar-improving, line 581: Can we please add validations for the fields.

```
- ['window_size_ms' => -1] → silently wraps to 4294967295 (uint32 overflow)
- ['window_size_ms' => 0] → passes through as 0
- ['failure_rate_threshold' => 2.0] → accepted, but a 200% failure rate makes no sense
- ['failure_rate_threshold' => 0.0] → accepted, but the circuit would never trip
- ['min_errors' => -5] → wraps to a uint32
- ['consecutive_successes' => 0] → accepted
```

Python adds these checks:
```
if window_size_ms <= 0:
    raise ValueError("window_size_ms must be positive")
if not (0.0 < failure_rate_threshold <= 1.0):
    raise ValueError("failure_rate_threshold must be between 0.0 (exclusive) and 1.0 (inclusive)")
if min_errors <= 0:
    raise ValueError("min_errors must be positive")
if open_timeout_ms <= 0:
    raise ValueError("open_timeout_ms must be positive")
if consecutive_successes <= 0:
    raise ValueError("consecutive_successes must be positive")
```

common.h
```diff
@@ -145,6 +145,31 @@ typedef struct {
...
+#define VALKEY_GLIDE_CB_COUNT_TIMEOUTS "count_timeouts"
+#define VALKEY_GLIDE_CB_CONSECUTIVE_SUCCESSES "consecutive_successes"
+
+#define VALKEY_GLIDE_CB_DEFAULT_WINDOW_SIZE_MS 10000
+#define VALKEY_GLIDE_CB_DEFAULT_FAILURE_RATE_THRESHOLD 0.5
+#define VALKEY_GLIDE_CB_DEFAULT_MIN_ERRORS 50
+#define VALKEY_GLIDE_CB_DEFAULT_OPEN_TIMEOUT_MS 5000
+#define VALKEY_GLIDE_CB_DEFAULT_COUNT_TIMEOUTS false
+#define VALKEY_GLIDE_CB_DEFAULT_CONSECUTIVE_SUCCESSES 3
+#define VALKEY_GLIDE_ERROR_TYPE_CIRCUIT_BREAKER_OPEN 4
```
currantw, line 162: Should we define/populate circuit breaker defaults at the client level – or should we let the Rust core set them? I am doing the latter on the C# client, so want to make sure these are aligned!

valkey_glide_ft_common.c
```diff
@@ -53,8 +53,9 @@ int execute_ft_command_internal(const void*      glide_client,
         return 0;
     }
     if (result->command_error) {
-        zend_throw_exception(
-            get_valkey_glide_exception_ce(), result->command_error->command_error_message, 0);
+        zend_throw_exception(get_exception_ce_for_command_error(result->command_error),
+                             result->command_error->command_error_message,
+                             0);
```
currantw: Since `get_exception_ce_for_command_error` seems to only be used in this format, would it make sense to just wrap this all into something like `throw_exception_for_command_error()`?

tests/ConnectionRequestTest.php
currantw, line 749: It seems like you have only added unit tests, not any integration tests. Do you think that is sufficient? I tried to add some basic integration tests (basically just that connection works, that the circuit breaker could be tripped with low thresholds) for C#.

currantw (commented):
Reviewed to 7ba5343.

tests/ConnectionRequestTest.php
jeremyprime: There are no edge case tests for any invalid options in any of the clients.  If we add them it will be a cross-client effort (see https://github.com/valkey-io/valkey-glide/issues/5937), not as part of this PR.

valkey_glide_commands_common.h
jeremyprime: There are 2 types of `zend_throw_exception` here, FFI command errors (which have `command_error_type`) and local validation errors which are thrown before reaching the Rust core. All FFI error paths use  `get_exception_ce_for_command_error`. The remaining ones are local and have no `CommandError` and can never return `CircuitBreakerOpen`.

valkey_glide.c
jeremyprime: As mentioned above, since no other config options have edge case validation I would rather see this done across all clients as part of https://github.com/valkey-io/valkey-glide/issues/5937.

common.h
jeremyprime: I think we should define and send from client (Java, Node, Python, Go already doing this too).  This reduces ambiguity and documents from client side what is expected, although does duplicate the values in core and client.  This is also how other features are handled (reconnection, compression, etc).

tests/ConnectionRequestTest.php
jeremyprime: Only Java has CB integration tests, other clients just have unit tests for CB.  I did this because CB is a bit more complex and I thought it made more sense to test the tripping and other scenarios as system tests (in our internal system test suite) instead.  Especially since some of those scenarios can be a bit flaky in an integration test.

valkey_glide_ft_common.c
jeremyprime: Done, now uses a common `throw_command_error`.

prateek-kumar-improving (approved):

valkey_glide.c
currantw: Sounds good to me. I will update C# accordingly.

common.h
currantw: Sounds good. I agree that the visibility of default values to the user is worth the duplication.

valkey_glide_ft_common.c
currantw: Thank you 🙏

tests/ConnectionRequestTest.php
currantw: Sounds good. I will update C# client accordingly.

currantw (approved):
Reviewed to cec7239.
Looks good ✅ 
Thanks for replying to all my comments.
Will update C# GLIDE client to use a consistent approach.

PR #282, opened 2026-07-27
Implement failover and replicaof command

prateek-kumar-improving:
### Summary

Implement REPLICAOF and FAILOVER commands for the standalone ValkeyGlide client.
### Issue link

- replicaof(?string $host = null, int $port = 6379) — Makes the server a replica of the specified primary. When called with null host, sends REPLICAOF NO ONE to promote the server to a primary.
- failover(?array $to = null, bool $abort = false, int $timeout = 0) — Starts a coordinated failover from the connected primary to one of its replicas. Supports TO host port [FORCE], ABORT, and TIMEOUT options.
- Both commands support OPT_REPLY_LITERAL (returns "OK" string instead of true).
- Standalone-only commands — not exposed on ValkeyGlideCluster (matching PHPRedis behavior).

### Implementation

- Added execute_failover_command() and execute_replicaof_command() in valkey_glide_commands.c
- Both use STANDARD_METHOD_IMPL macro pattern
- Commands registered as FailOver (1127) and ReplicaOf (1151) request types in glide-core
- Added to prepare_message_args switch case in valkey_glide_core_common.c (alphabetized)
- Added independent Valkey server on port 6382 in CI for isolated testing

### Limitations

- FAILOVER success test performs best-effort topology restoration after failover completes. The glide client connection may become unstable after a role change.
- Cluster mode does not support these commands (standalone-only).

### Testing

All tests are skipped in cluster mode via overrides in ValkeyGlideClusterTest.

### Checklist


tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+            addresses: [['host' => '127.0.0.1', 'port' => 6382]]
+        );
+
+        // Verify 6382 is a master with connected replicas
+        $info = $client->info('REPLICATION');
+        $this->assertEquals('master', $info['role']);
+        $this->assertGTE(1, (int) $info['connected_slaves']);
+
+        // FAILOVER returns true (command accepted, failover is async)
+        $result = $client->failover();
```
nderraugh: Does anything need to get reset with the servers after this test succeeds or fails after an attempted failover so that other following tests aren't affected? i.e. something to run in a tear down-like phase to reset the servers.

valkey_glide_commands.c
```diff
@@ -720,6 +720,144 @@ int execute_client_unpause_command(zval*             object,
...
+
+    /* Setup core command arguments */
+    core_command_args_t core_args = {0};
+    core_args.glide_client        = valkey_glide->glide_client;
+    core_args.cmd_type            = FailOver;
+    core_args.is_cluster          = 0;
+
+    int arg_idx = 0;
+
+    if (abort) {
```
nderraugh, line 766: `abort` doesn't use the timeout parameter?

nderraugh: If these parameters (host, abort, timeout) have conflicting permutations, it might be worth addressing them the same way it's done in other places.

i.e. `@throws ValkeyGlideException If conflicting parameters are specified or connection fails`.

Alternatively succeeding and ignoring the conflict might be worth logging.  Or perhaps even splitting functions along the abort path.

valkey_glide_commands.c
```diff
@@ -720,6 +720,144 @@ int execute_client_unpause_command(zval*             object,
...
+        core_args.args[arg_idx].data.string_arg.value = "ABORT";
+        core_args.args[arg_idx].data.string_arg.len   = 5;
+        arg_idx++;
+    } else {
+        if (to_arr) {
+            /* Extract host and port from array */
+            zval* z_host = zend_hash_str_find(Z_ARRVAL_P(to_arr), "host", 4);
+            zval* z_port = zend_hash_str_find(Z_ARRVAL_P(to_arr), "port", 4);
+
+            if (!z_host || Z_TYPE_P(z_host) != IS_STRING || !z_port) {
```
nderraugh: Similar issue around validation here.  

Existing validation patterns look like this
```suggestion
            if (!z_host || Z_TYPE_P(z_host) != IS_STRING || !z_port) {
    VALKEY_LOG_ERROR("command_validation", "'to' array must contain string 'host' and 'port'");
    zend_throw_exception(get_valkey_glide_exception_ce(),
                         "'to' array must contain string 'host' and 'port'", 0);
```

I think it makes sense to use that here.  Otherwise a malformed `$to` fails silently with no hint about the problem.

nderraugh: Might be worth thinking about whether this can be done cohesively around lines 737-740.

valkey_glide_commands.c
```diff
@@ -720,6 +720,144 @@ int execute_client_unpause_command(zval*             object,
...
+        /* REPLICAOF NO ONE */
+        core_args.args[0].type                  = CORE_ARG_TYPE_STRING;
+        core_args.args[0].data.string_arg.value = "NO";
+        core_args.args[0].data.string_arg.len   = 2;
+        core_args.args[1].type                  = CORE_ARG_TYPE_STRING;
+        core_args.args[1].data.string_arg.value = "ONE";
+        core_args.args[1].data.string_arg.len   = 3;
+        core_args.arg_count                     = 2;
+    } else {
+        /* REPLICAOF host port */
```
nderraugh, line 875: Do we need to do any validation on `host` here?  e.g. should '' be normalized or rejected (and logged iwth explanation) or something else?

valkey_glide.stub.php
```diff
@@ -2438,6 +2438,39 @@ public function clientUnpause(): ValkeyGlide|bool|string;
...
+     *                                  false on failure.
+     *
+     * @see https://valkey.io/commands/replicaof
+     */
+    public function replicaof(?string $host = null, int $port = 6379): ValkeyGlide|bool|string;
+
+    /**
+     * Starts a coordinated failover from the connected primary to one of its replicas.
+     *
+     * @param array|null $to    Target replica as ['host' => string, 'port' => int], or null for auto-selection.
```
currantw: Nit. 
```suggestion
     * @param array|null $to Target replica as ['host' => string, 'port' => int], or null for auto-selection.
```

valkey_z_php_methods.c
```diff
@@ -814,14 +814,22 @@ FLUSHDB_METHOD_IMPL(ValkeyGlide)
 FLUSHALL_METHOD_IMPL(ValkeyGlide)
 /* }}} */
 
-/* {{{ proto string ValkeyGlide::bgSave([string mode]) */
+/* {{{ proto bool ValkeyGlide::bgSave([string mode]) */
```
currantw, line 817: Thanks for fixing 💪

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+    {
+        // Use independent server on port 6382 (standalone primary, no replicas)
+        $client = new ValkeyGlide();
+        $client->connect(
+            addresses: [['host' => '127.0.0.1', 'port' => 6382]]
+        );
+
+        // Verify it's currently a master
+        $info = $client->info('REPLICATION');
+        $this->assertEquals('master', $info['role']);
```
currantw: Can we use `waitForRole` here?

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+        // Wait for replication to establish and verify role changed
+        $this->waitForRole($client, 'slave');
+
+        // Promote back to primary with REPLICAOF NO ONE
+        $result = $client->replicaof();
+        $this->assertTrue($result);
+
+        // Verify role changed back to master
+        $info = $client->info('REPLICATION');
+        $this->assertEquals('master', $info['role']);
```
currantw: Can we use `waitForRole` here?

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+            // Wait for replication to establish
+            $client->setOption(ValkeyGlide::OPT_REPLY_LITERAL, false);
+            $this->waitForRole($client, 'slave');
+            $client->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
+
+            $result = $client->replicaof();
+            $this->assertEquals('OK', $result);
+        } finally {
+            $client->setOption(ValkeyGlide::OPT_REPLY_LITERAL, false);
+        }
```
currantw: Didn't we add a helper to simplify the `OPT_REPLY_LITERAL` cases?

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+
+        // Verify role restored to master
+        $info = $client->info('REPLICATION');
+        $this->assertEquals('master', $info['role']);
+
```
currantw: Can we use `waitForRole` here?

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+            addresses: [['host' => '127.0.0.1', 'port' => 6384]]
+        );
+
+        $client->setOption(ValkeyGlide::OPT_REPLY_LITERAL, true);
+        try {
+            $result = $client->failover();
+            $this->assertFalse($result);
+        } finally {
+            $client->setOption(ValkeyGlide::OPT_REPLY_LITERAL, false);
+        }
```
currantw: Similar comment about using `OPT_REPLY_LITERAL` helper.

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+        // Use isolated server on port 6382 which has 6383 as its replica.
+        // This avoids disrupting the shared 6379 primary used by other tests.
+        $client = new ValkeyGlide();
+        $client->connect(
+            addresses: [['host' => '127.0.0.1', 'port' => 6382]]
+        );
+
+        // Verify 6382 is a master with connected replicas
+        $info = $client->info('REPLICATION');
+        $this->assertEquals('master', $info['role']);
```
currantw: Use `waitForRole` here?

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3287,160 @@ public function testPersist()
...
+        $replica->replicaof('127.0.0.1', 6382);
+        $replica->close();
+
+        $client->close();
+    }
+
+    /**
+     * Poll until a server reaches the expected role, or fail after timeout.
+     */
+    protected function waitForRole(ValkeyGlide $client, string $expectedRole, int $timeoutMs = 5000): void
```
currantw: Do we have a generic `waitFor` base method that this could use? We have something similar in the other clients. If not, I think it makes sense to add it?

currantw: We should also add an `assertRole` method. This could be used for cases where we don't need to wait _and_ passed to `waitFor` for the case where we do?

tests/ValkeyGlideTest.php
currantw: (Or `assertRole` comments suggested in other comment)

currantw (commented):
Reviewed to 49822f9.

tests/ValkeyGlideTest.php
currantw: (Or `assertRole` comments suggested in other comment)

tests/ValkeyGlideTest.php
currantw: (Or `assertRole` comments suggested in other comment)

tests/ValkeyGlideTest.php
currantw: (Or `assertRole` comments suggested in other comment)

valkey_glide.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Yes, rebased from main. Adding to this branch.
Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
prateek-kumar-improving: Wrapped the failover assertion in a try/finally block that always restores 6382 to primary and re-attaches 6383 as its replica. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

valkey_glide_commands.c
prateek-kumar-improving: FAILOVER ABORT takes no additional arguments. Added validation that throws ValkeyGlideException if abort is combined with $to or $timeout to make this explicit. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

prateek-kumar-improving: Added early validation that rejects conflicting parameter combinations with VALKEY_LOG_ERROR + zend_throw_exception, and documented with `@throws` in the stub.

prateek-kumar-improving: Fixed. Added VALKEY_LOG_ERROR and zend_throw_exception following the existing pattern. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

valkey_glide_commands.c
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

valkey_glide_commands.c
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/5f6866a5922c124404c2594cb8e76fc5a77db8c6

tests/ValkeyGlideTest.php
```diff
@@ -114,14 +114,18 @@ protected function getMigrateDestClient(): ValkeyGlide
     /**
      * Execute a callable with OPT_REPLY_LITERAL enabled, ensuring it is
      * always disabled afterwards even if an exception is thrown.
+     *
+     * @param callable        $fn     The callable to execute with OPT_REPLY_LITERAL enabled.
+     * @param ValkeyGlide|null $client Optional client instance. Defaults to $this->valkey_glide.
```
currantw: Nit. Is this something you need to tweak your agent on perhaps? 🤔 


```suggestion
     * @param callable         $fn The callable to execute with OPT_REPLY_LITERAL enabled.
     * @param ValkeyGlide|null $client Optional client instance. Defaults to $this->valkey_glide.
```

tests/ValkeyGlideTest.php
```diff
@@ -3287,6 +3291,162 @@ public function testPersist()
...
+    {
+        $info = $client->info('REPLICATION');
+        $this->assertIsArray($info);
+        $this->assertEquals($expectedRole, $info['role']);
+    }
+
+    /**
+     * Poll until a server reaches the expected role, or fail after timeout.
+     */
+    protected function waitForRole(ValkeyGlide $client, string $expectedRole, int $timeoutMs = 5000): void
```
currantw: Can we remove `$timeoutMs` – it isn't used, is it?

valkey_glide_commands.c
nderraugh: Are host, and other similar parameters, validated in rust-core?

valkey_glide_commands.c
```diff
@@ -720,6 +720,174 @@ int execute_client_unpause_command(zval*             object,
...
+    valkey_glide = VALKEY_GLIDE_PHP_ZVAL_GET_OBJECT(valkey_glide_object, object);
+    if (!valkey_glide || !valkey_glide->glide_client) {
+        return 0;
+    }
+
+    /* Parse parameters: failover(?array $to = null, bool $abort = false, int $timeout = 0) */
+    if (zend_parse_method_parameters(
+            argc, object, "O|a!bl", &object, ce, &to_arr, &abort, &timeout) == FAILURE) {
+        return 0;
+    }
```
currantw, line 740: How does timeout = 0 get handled? It seems like this might silently default to "no timeout", which probably isn't the behaviour I would expect?

currantw (commented):
Reviewed to 0a6f677.

tests/ValkeyGlideTest.php
nderraugh: If that try fails and nothing is caught (and asserted as a failure) will this test look like it passed?

tests/ValkeyGlideTest.php
prateek-kumar-improving: It is used here
```
protected function waitForRole(ValkeyGlide $client, string $expectedRole, int $timeoutSeconds = 5): void
{
    $this->waitFor(
        function () use ($client, $expectedRole) {
            $info = $client->info('REPLICATION');
            return is_array($info) && ($info['role'] ?? null) === $expectedRole;
        },
        $timeoutSeconds,  // <-- passed here as 2nd arg
        "Timed out waiting for role to become '$expectedRole'"
    );
}
```
Changing the name to timeoutSeconds.

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/e50e64a2dbf77b86934974a4e419fd73b5587238

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/282/changes/e50e64a2dbf77b86934974a4e419fd73b5587238

valkey_glide_commands.c
prateek-kumar-improving: Correct — timeout = 0 means no TIMEOUT argument is sent to the server, which per the Valkey FAILOVER spec means the failover has no deadline. This matches PHPRedis behavior and the Valkey docs (TIMEOUT milliseconds is optional). The parameter default of 0 is documented as 'no timeout' in the docblock.

tests/ValkeyGlideTest.php
prateek-kumar-improving: if the try block throws (assertion failure or exception), the exception propagates after the finally block executes. So the test still fails with the original error, but cleanup is guaranteed to run. PHP's finally doesn't swallow exceptions.

valkey_glide_commands.c
prateek-kumar-improving: We don't validate them. For other commands also we don't validate these. We will let the server throw the error.
We validate in very specific cases. But in most cases we just let server throw the error.

nderraugh (approved):

valkey_glide_commands.c
currantw: Great thanks!

tests/ValkeyGlideTest.php
currantw: The timeout for `waitFor` is optional. Why are we provided 5 seconds here instead of just (a) removing the `timeout` parameter completely from `waitForRole` and so (b) just falling back to the default value? I don't think we are ever actually calling `waitForRole` with a different timeout, are we? Adding optional parameters when there is no use case just increases the complexity of our code base 🤷

currantw (commented):
Reviewed to e50e64a.

tests/ValkeyGlideTest.php
prateek-kumar-improving: These are generic methods that should be written in a way that they can be reused for other cases as well. Providing an extra variable just provides additional support if someone needs to update the timeout.

currantw: I disagree. Unless there is a clear reason for adding additional parameters (i.e. a scheduled issue, a known bug, etc), I don't think we should do so. If it is needed in the future, it is easy for a future developer to add. If we add it now, it adds additional complexity, maintenance, and documentation with no benefit.

In all likelihood, this will never get used – I have a very hard time imagining any scenario where a test would need to call `waitForRole` with a custom timeout that is different from every other test that uses that method – but it will _forever_ add extra bloat to our code base, whether or not it is needed.

Every individual extra "just in case" parameter represents a minimal amount of additional complexity, but complexity is non-linear: when we add up dozens or hundreds of such cases, it can be very difficult to actually understand that code base. 

That being said, I will defer to you. Please resolve once read.

currantw (approved):
Reviewed to e50e64a.
Looks good.
One comment to resolve once read.

PR #283, opened 2026-07-28
Implement LATENCY HISTORY, LATEST, and RESET commands

prateek-kumar-improving:
### Summary

Implement LATENCY HISTORY, LATENCY LATEST, and LATENCY RESET commands for both standalone and cluster clients.

### Issue link

- latencyHistory(string $event): array|false — Returns latency spike time series for the specified event as an array of [timestamp, duration_ms] pairs. Returns 
empty array if event does not exist.
- latencyLatest(): array|false — Reports the latest latency events as an array of [event_name, timestamp, duration_ms, max_duration_ms] entries.
- latencyReset(string ...$events): int|false — Resets latency data for specified events (or all events if none specified). Returns the number of events reset.
- All three commands are available on both standalone (ValkeyGlide) and cluster (ValkeyGlideCluster) clients.

### Implementation

- Added execute_latency_history_command(), execute_latency_latest_command(), and execute_latency_reset_command() in valkey_glide_commands.c
- All use STANDARD_METHOD_IMPL macro pattern
- LatencyLatest registered as zero-arg operation, LatencyHistory/LatencyReset as message operations in valkey_glide_core_common.c
- Uses process_core_array_result for history/latest and process_core_int_result for reset
- Cluster stubs use @see only per review conventions

### Limitations

- PHPRedis does not have dedicated latency methods, so there is no PHPRedis signature to match. The API follows Valkey command semantics directly.
- Tests use rawCommand for CONFIG GET/SET to trigger latency spikes, as config() is not yet available on the cluster client. Issue link: https://github.com/valkey-io/valkey-glide-php/issues/284

### Testing

Tests run in both standalone and cluster mode (inherited).

### Checklist

### DIFF
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -900,6 +900,12 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $result = $this->valkey_glide->latencyHistory('command');
         $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
+
+        // Non-existent event returns empty array
+        $unknown = $this->valkey_glide->latencyHistory('nonexistent_event');
+        $this->assertIsArray($unknown);
+        $this->assertEquals(0, count($unknown));
     }
 
     public function testLatencyLatest()
@@ -908,14 +914,24 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         $result = $this->valkey_glide->latencyLatest();
         $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
     }
 
     public function testLatencyReset()
     {
         $this->triggerLatencySpike();
 
+        // Verify history exists before reset
+        $history = $this->valkey_glide->latencyHistory('command');
+        $this->assertGT(0, count($history));
+
         $result = $this->valkey_glide->latencyReset();
-        $this->assertTrue(is_int($result) || is_array($result));
+        $this->assertIsInt($result);
+        $this->assertGT(0, $result);
+
+        // History should be empty after reset
+        $history = $this->valkey_glide->latencyHistory('command');
+        $this->assertEquals(0, count($history));
     }
 
     public function testLatencyResetSpecificEvent()
@@ -923,15 +939,26 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->triggerLatencySpike();
 
         $result = $this->valkey_glide->latencyReset('command');
-        $this->assertTrue(is_int($result) || is_array($result));
+        $this->assertIsInt($result);
+        $this->assertGT(0, $result);
+
+        // History for "command" should be empty
+        $history = $this->valkey_glide->latencyHistory('command');
+        $this->assertEquals(0, count($history));
     }
 
     public function testLatencyResetUnknownEvent()
     {
         $this->triggerLatencySpike();
 
+        // Reset unknown event returns 0
         $result = $this->valkey_glide->latencyReset('unknown_event');
-        $this->assertTrue(is_int($result) || is_array($result));
+        $this->assertIsInt($result);
+        $this->assertEquals(0, $result);
+
+        // History for "command" should still exist
+        $history = $this->valkey_glide->latencyHistory('command');
+        $this->assertGT(0, count($history));
     }
 
     public function testLatencyHistoryWithRoute()
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -1042,6 +1042,26 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertIsInt($result);
     }
 
+    public function testLatencyHistoryWithNullRoute()
+    {
+        $this->triggerLatencySpike();
+
+        // Explicit null should behave like an omitted route
+        $result = $this->valkey_glide->latencyHistory('command', null);
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
+    }
+
+    public function testLatencyLatestWithNullRoute()
+    {
+        $this->triggerLatencySpike();
+
+        // Explicit null should behave like an omitted route
+        $result = $this->valkey_glide->latencyLatest(null);
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
+    }
+
     public function testReset()
     {
         $key = '{reset_test}_' . uniqid();
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -907,7 +907,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
         try {
             // DEBUG SLEEP on a random node — all nodes have threshold=1 so the spike is recorded
-            $this->valkey_glide->rawCommand('', 'DEBUG', 'SLEEP', '0.05');
+            $this->valkey_glide->rawCommand('randomNode', 'DEBUG', 'SLEEP', '0.05');
         } finally {
             $this->valkey_glide->rawCommand('allPrimaries', 'CONFIG', 'SET', 'latency-monitor-threshold', $prevThreshold);
         }
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -977,6 +977,14 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertIsArray($result);
     }
 
+    public function testLatencyResetWithRoute()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyResetWithRoute('randomNode');
+        $this->assertIsInt($result);
+    }
+
     public function testReset()
     {
         $key = '{reset_test}_' . uniqid();
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -1022,24 +1022,42 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $this->triggerLatencySpike();
 
+        // randomNode returns a flat array of [timestamp, duration] pairs from one node
         $result = $this->valkey_glide->latencyHistory('command', 'randomNode');
         $this->assertIsArray($result);
+        // Verify entries have correct shape if any exist
+        foreach ($result as $entry) {
+            $this->assertIsArray($entry);
+            $this->assertEquals(2, count($entry));
+            $this->assertGT(0, $entry[0]); // timestamp > 0
+            $this->assertGT(0, $entry[1]); // duration > 0
+        }
     }
 
     public function testLatencyLatestWithRoute()
     {
         $this->triggerLatencySpike();
 
+        // randomNode returns a flat array of latest events from one node
         $result = $this->valkey_glide->latencyLatest('randomNode');
         $this->assertIsArray($result);
+        // Verify entries have correct shape if any exist
+        foreach ($result as $entry) {
+            $this->assertIsArray($entry);
+            $this->assertGTE(4, count($entry));
+            $this->assertIsString($entry[0]); // event name
+            $this->assertGT(0, $entry[1]);    // timestamp > 0
+        }
     }
 
     public function testLatencyResetWithRoute()
     {
         $this->triggerLatencySpike();
 
+        // randomNode returns an integer count from one node
         $result = $this->valkey_glide->latencyResetWithRoute('randomNode');
         $this->assertIsInt($result);
+        $this->assertGTE(0, $result);
     }
 
     public function testLatencyHistoryWithNullRoute()
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -896,18 +896,20 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     protected function triggerLatencySpike()
     {
-        // Cluster rawCommand requires route as first arg
-        $config = $this->valkey_glide->rawCommand('', 'CONFIG', 'GET', 'latency-monitor-threshold');
+        // Get current threshold from one node (all should have same value)
+        $config = $this->valkey_glide->rawCommand('randomNode', 'CONFIG', 'GET', 'latency-monitor-threshold');
         $prevThreshold = $config['latency-monitor-threshold'] ?? '0';
 
         $this->valkey_glide->latencyReset();
 
-        $this->valkey_glide->rawCommand('', 'CONFIG', 'SET', 'latency-monitor-threshold', '1');
+        // Set threshold on all primaries so any node records spikes
+        $this->valkey_glide->rawCommand('allPrimaries', 'CONFIG', 'SET', 'latency-monitor-threshold', '1');
 
         try {
+            // DEBUG SLEEP on a random node — all nodes have threshold=1 so the spike is recorded
             $this->valkey_glide->rawCommand('', 'DEBUG', 'SLEEP', '0.05');
         } finally {
-            $this->valkey_glide->rawCommand('', 'CONFIG', 'SET', 'latency-monitor-threshold', $prevThreshold);
+            $this->valkey_glide->rawCommand('allPrimaries', 'CONFIG', 'SET', 'latency-monitor-threshold', $prevThreshold);
         }
     }
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -894,6 +894,23 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertGT(0, $result['peak.allocated']);
     }
 
+    protected function triggerLatencySpike()
+    {
+        // Cluster rawCommand requires route as first arg
+        $config = $this->valkey_glide->rawCommand('', 'CONFIG', 'GET', 'latency-monitor-threshold');
+        $prevThreshold = $config['latency-monitor-threshold'] ?? '0';
+
+        $this->valkey_glide->latencyReset();
+
+        $this->valkey_glide->rawCommand('', 'CONFIG', 'SET', 'latency-monitor-threshold', '1');
+
+        try {
+            $this->valkey_glide->rawCommand('', 'DEBUG', 'SLEEP', '0.05');
+        } finally {
+            $this->valkey_glide->rawCommand('', 'CONFIG', 'SET', 'latency-monitor-threshold', $prevThreshold);
+        }
+    }
+
     public function testLatencyHistory()
     {
         $this->triggerLatencySpike();
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -917,12 +917,12 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $this->triggerLatencySpike();
 
-        $result = $this->valkey_glide->latencyHistory('command');
+        $result = $this->valkey_glide->latencyHistory('command', 'randomNode');
         $this->assertIsArray($result);
         $this->assertGT(0, count($result));
 
         // Non-existent event returns empty array
-        $unknown = $this->valkey_glide->latencyHistory('nonexistent_event');
+        $unknown = $this->valkey_glide->latencyHistory('nonexistent_event', 'randomNode');
         $this->assertIsArray($unknown);
         $this->assertEquals(0, count($unknown));
     }
@@ -940,8 +940,8 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
     {
         $this->triggerLatencySpike();
 
-        // Verify history exists before reset
-        $history = $this->valkey_glide->latencyHistory('command');
+        // Verify history exists before reset (use randomNode for scalar response)
+        $history = $this->valkey_glide->latencyHistory('command', 'randomNode');
         $this->assertGT(0, count($history));
 
         $result = $this->valkey_glide->latencyReset();
@@ -949,7 +949,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertGT(0, $result);
 
         // History should be empty after reset
-        $history = $this->valkey_glide->latencyHistory('command');
+        $history = $this->valkey_glide->latencyHistory('command', 'randomNode');
         $this->assertEquals(0, count($history));
     }
 
@@ -962,7 +962,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertGT(0, $result);
 
         // History for "command" should be empty
-        $history = $this->valkey_glide->latencyHistory('command');
+        $history = $this->valkey_glide->latencyHistory('command', 'randomNode');
         $this->assertEquals(0, count($history));
     }
 
@@ -976,7 +976,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertEquals(0, $result);
 
         // History for "command" should still exist
-        $history = $this->valkey_glide->latencyHistory('command');
+        $history = $this->valkey_glide->latencyHistory('command', 'randomNode');
         $this->assertGT(0, count($history));
     }
 
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -894,6 +894,62 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertGT(0, $result['peak.allocated']);
     }
 
+    public function testLatencyHistory()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyHistory('command');
+        $this->assertIsArray($result);
+    }
+
+    public function testLatencyLatest()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyLatest();
+        $this->assertIsArray($result);
+    }
+
+    public function testLatencyReset()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyReset();
+        $this->assertTrue(is_int($result) || is_array($result));
+    }
+
+    public function testLatencyResetSpecificEvent()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyReset('command');
+        $this->assertTrue(is_int($result) || is_array($result));
+    }
+
+    public function testLatencyResetUnknownEvent()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyReset('unknown_event');
+        $this->assertTrue(is_int($result) || is_array($result));
+    }
+
+    public function testLatencyHistoryWithRoute()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyHistory('command', 'randomNode');
+        $this->assertIsArray($result);
+    }
+
+    public function testLatencyLatestWithRoute()
+    {
+        $this->triggerLatencySpike();
+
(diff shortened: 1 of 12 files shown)


valkey_glide_commands.c
```diff
@@ -610,6 +610,126 @@ int execute_save_command(zval* object, int argc, zval* return_value, zend_class_
...
+            if (Z_TYPE(args[i]) == IS_STRING) {
+                all[arg_idx].type                  = CORE_ARG_TYPE_STRING;
+                all[arg_idx].data.string_arg.value = Z_STRVAL(args[i]);
+                all[arg_idx].data.string_arg.len   = Z_STRLEN(args[i]);
+                arg_idx++;
+            }
+        }
+        core_args.all_args = all;
+    }
+    core_args.arg_count = arg_idx;
```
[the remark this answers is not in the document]
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/commits/77a39a7

tests/ValkeyGlideTest.php
```diff
@@ -3260,6 +3260,107 @@ public function testMemoryStats()
         $this->valkey_glide->del($key);
     }
 
+    // TODO: Replace rawCommand CONFIG calls with config() once config() is available on cluster client.
+    // See: https://github.com/valkey-io/valkey-glide-php/issues/284
+    protected function triggerLatencySpike()
+    {
+        // Save current threshold
+        $config = $this->valkey_glide->rawCommand('', 'CONFIG', 'GET', 'latency-monitor-threshold');
```
nderraugh: In standalone isn't the first parameter the command?

valkey_glide_commands.c
```diff
@@ -620,6 +620,226 @@ static int process_memory_stats_result(CommandResponse* response,
...
+            all[arg_idx].data.string_arg.value = Z_STRVAL(events[i]);
+            all[arg_idx].data.string_arg.len   = Z_STRLEN(events[i]);
+            arg_idx++;
+        }
+        core_args->all_args = all;
+    }
+    core_args->arg_count = arg_idx;
+
+    int result = execute_and_handle_batch(
+        valkey_glide, core_args, process_core_int_result, return_value, object);
```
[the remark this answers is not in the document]
prateek-kumar-improving: - LATENCY HISTORY / LATENCY LATEST have ResponsePolicy::Special which glide-core handles internally
- LATENCY RESET has ResponsePolicy::Aggregate(AggregateOp::Sum) which sums to a scalar

tests/ValkeyGlideTest.php
prateek-kumar-improving: yes, it is command for standalone and for cluster it is route. This method is used by both standalone and cluster and raw command just concatenates '' to the start of command. Therefore, it works. Since raw command in standalone will just add an empty string before CONFIG and it will still run.

But to avoid confusion updating it here: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/ecfb9fb545453e8323c2afb3a0d249382b43fca4

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,114 @@ public function testMemoryStatsWithRoute()
...
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyReset('command');
+        $this->assertIsInt($result);
+        $this->assertGT(0, $result);
+
+        // History for "command" should be empty
+        $history = $this->valkey_glide->latencyHistory('command');
+        $this->assertEquals(0, count($history));
```
xShinnRyuu: `testLatencyResetSpecificEvent` looks racy in cluster mode. `triggerLatencySpike` runs `DEBUG SLEEP` through `rawCommand('', ...)`, which routes randomly, so the `command` latency record lands on one specific primary. `latencyReset('command')` has no route argument here, so it relies on glide-core's default aggregate routing to clear every node's record. The follow-up `latencyHistory('command')` also runs with no route and hits a single random primary. If that primary never saw a spike in this iteration, `count($history)` is already 0 whether reset worked or not. Two options: pin the spike, the reset, and the read to the same primary with an explicit route, or assert on the returned reset count instead of on the post-reset history.

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,114 @@ public function testMemoryStatsWithRoute()
...
+        $this->triggerLatencySpike();
+
+        // Reset unknown event returns 0
+        $result = $this->valkey_glide->latencyReset('unknown_event');
+        $this->assertIsInt($result);
+        $this->assertEquals(0, $result);
+
+        // History for "command" should still exist
+        $history = $this->valkey_glide->latencyHistory('command');
+        $this->assertGT(0, count($history));
```
xShinnRyuu: Same routing concern as `testLatencyResetSpecificEvent`. This test asserts that history for `command` still exists after resetting `unknown_event`, but the `DEBUG SLEEP` that seeded `command` hit one random primary, and the follow-up `latencyHistory('command')` also hits one random primary. When those are different nodes, `assertGT(0, count($history))` is really testing whether an unrelated primary happened to record a `command` spike, not what the test name implies. Pinning both the spike and the read to the same node with an explicit route would make this deterministic.

valkey_glide_commands.c
```diff
@@ -620,6 +620,226 @@ static int process_memory_stats_result(CommandResponse* response,
...
+
+static int latency_reset_build_and_execute(valkey_glide_object* valkey_glide,
+                                           core_command_args_t* core_args,
+                                           zval*                events,
+                                           int                  event_count,
+                                           zval*                return_value,
+                                           zval*                object) {
+    int arg_idx = 0;
+
+    if (event_count <= 12) {
```
xShinnRyuu: The `12` threshold here is the size of `core_command_args_t.args` in `valkey_glide_core_common.h` (`core_arg_t args[12]`). Hard-coding the number in this branch means the two definitions can drift silently. If the fixed array shrinks, the `<=12` inline path overflows the stack; if it grows, the `ecalloc` branch fires earlier than it needs to. A shared constant on the struct definition (something like a `CORE_ARGS_MAX` macro), used here as well, would keep them in sync.

valkey_glide_cluster.stub.php
```diff
@@ -1035,6 +1035,36 @@ public function memoryPurge(mixed $route = null): ValkeyGlideCluster|bool|array|
...
+    /**
+     * Resets latency data with explicit routing.
+     *
+     * @param mixed  $route  Routing: 'allPrimaries', 'allNodes', 'randomNode', or a specific node address.
+     *
+     * @return ValkeyGlideCluster|int|false Returns the number of events reset. Returns false on failure.
+     *
+     * @see https://valkey.io/commands/latency-reset
+     */
+    public function latencyResetWithRoute(mixed $route, string ...$events): ValkeyGlideCluster|int|false;
```
xShinnRyuu, line 1087: `latencyHistory` and `latencyLatest` take the route as an optional trailing parameter on the same method, but `latencyReset` splits it into a separate `latencyResetWithRoute`. Callers who want a route now have to remember which of two method names applies, and both paths land on the same C implementation anyway. Would it be simpler to give `latencyReset` an optional trailing `$route` (or accept it first, the way `latencyResetWithRoute` does) and drop the second method? Not blocking, more of a consistency question.

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,114 @@ public function testMemoryStatsWithRoute()
...
 
+    protected function triggerLatencySpike()
+    {
+        // Cluster rawCommand requires route as first arg
+        $config = $this->valkey_glide->rawCommand('', 'CONFIG', 'GET', 'latency-monitor-threshold');
+        $prevThreshold = $config['latency-monitor-threshold'] ?? '0';
+
+        $this->valkey_glide->latencyReset();
+
+        $this->valkey_glide->rawCommand('', 'CONFIG', 'SET', 'latency-monitor-threshold', '1');
```
nderraugh: These empty routes look like they will fail to execute commands because the route length will be 0 and create_route_bytes_from_route will return NULL. Am I reading that right?

If that's correct then I don't think this method will trigger any latency.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/9b95cf1136fe594e567797c3612858afdf639633

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/9b95cf1136fe594e567797c3612858afdf639633

valkey_glide_commands.c
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/fe2798678f9928a1b78d44b960f258361d811fff

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated to randomNode in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/fe2798678f9928a1b78d44b960f258361d811fff

valkey_glide_cluster.stub.php
prateek-kumar-improving: We can't put it after variadic variable:
`public function latencyReset(string ...$events, mixed $route = null) // INVALID PHP syntax`

We need to keep consistency in syntax, so we will need to keep the `$events` variable first. A separate method with route is the only good option.

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,116 @@ public function testMemoryStatsWithRoute()
...
+        $prevThreshold = $config['latency-monitor-threshold'] ?? '0';
+
+        $this->valkey_glide->latencyReset();
+
+        // Set threshold on all primaries so any node records spikes
+        $this->valkey_glide->rawCommand('allPrimaries', 'CONFIG', 'SET', 'latency-monitor-threshold', '1');
+
+        try {
+            // DEBUG SLEEP on a random node — all nodes have threshold=1 so the spike is recorded
+            $this->valkey_glide->rawCommand('randomNode', 'DEBUG', 'SLEEP', '0.05');
```
nderraugh, line 910: I think practically speaking this test is correct, although I haven't checked the cluster setup.

But allPrimaries only changes the threshold on primary nodes. So DEBUG SLEEP could be sent to a replica whose threshold was never changed, producing no latency event.

So practicality aside we should fix this so that the intent of the test is clear.

prateek-kumar-improving:
```
if redis::cluster_routing::is_readonly_cmd(cmd_name.as_bytes()) {
    RoutingInfo::SingleNode(SingleNodeRoutingInfo::Random)
} else {
    // Writable command → RandomPrimary
    RoutingInfo::SingleNode(SingleNodeRoutingInfo::RandomPrimary)
}
```

DEBUG SLEEP is not a readonly command, so glide-core automatically promotes Random routing to RandomPrimary. The sleep always hits a primary that has the threshold set.

valkey_glide_commands.c
```diff
@@ -620,6 +620,226 @@ static int process_memory_stats_result(CommandResponse* response,
...
+        }
+    }
+
+    core_args.args[0].type                  = CORE_ARG_TYPE_STRING;
+    core_args.args[0].data.string_arg.value = event;
+    core_args.args[0].data.string_arg.len   = event_len;
+    core_args.arg_count                     = 1;
+
+    return execute_and_handle_batch(
+        valkey_glide, &core_args, process_core_array_result, return_value, object);
```
nderraugh: latencyHistory() and latencyLatest() always use process_core_array_result (valkey_glide_commands.c:672,715), which converts maps non-associatively. An allNodes or allPrimaries response becomes: [nodeAddress, nodeResult, nodeAddress, nodeResult, ...]
instead of a node-keyed map. 

latencyResetWithRoute('allNodes') uses process_core_int_result (valkey_glide_commands.c:746), which rejects a map response and returns false. The cluster stub explicitly advertises these fan-out routes, but tests cover only randomNode. 

So..

1.. Cluster latencyHistory() and latencyLatest() default to routing across all primaries.
2. GLIDE Core returns multi-node results as a map:
```php
[
    '127.0.0.1:7001' => $node1Result,
    '127.0.0.1:7002' => $node2Result,
    '127.0.0.1:7003' => $node3Result,
]
```
3. Both commands use process_core_array_result() in valkey_glide_commands.c.
4. That processor requests non-associative map conversion, flattening the map into:
```php
[
    '127.0.0.1:7001', $node1Result,
    '127.0.0.1:7002', $node2Result,
    '127.0.0.1:7003', $node3Result,
]
```

This loses the node-address-to-result relationship.

**CI confirmation**
Current CI reproduces it:

```
Assertion failed: testLatencyHistory - 6 !== 0
Assertion failed: testLatencyReset - 6 !== 0
Assertion failed: testLatencyResetSpecificEvent - 6 !== 0
```
The 6 is three primary-node entries flattened into six alternating key/value elements.

The RESET tests fail during their subsequent latencyHistory() verification.

**Tentative Fix**
Give HISTORY and LATEST a response processor that uses `COMMAND_RESPONSE_ASSOSIATIVE_ARRAY_MAP`.

This preserves node addresses for default, allPrimaries, and allNodes routing while leaving single-node native array responses intact.

Update cluster tests:
* Do not expect an empty top-level array after an unknown event or reset.
* Expect a node-addressed map whose per-node history arrays are empty.
* Add explicit allPrimaries and allNodes coverage. Existing randomNode tests exercise only the single-node response and cannot detect this bug.

valkey_glide_commands.c
```diff
@@ -620,6 +620,226 @@ static int process_memory_stats_result(CommandResponse* response,
...
+        core_arg_t* all = (core_arg_t*) ecalloc(event_count, sizeof(core_arg_t));
+        for (int i = 0; i < event_count; i++) {
+            all[arg_idx].type                  = CORE_ARG_TYPE_STRING;
+            all[arg_idx].data.string_arg.value = Z_STRVAL(events[i]);
+            all[arg_idx].data.string_arg.len   = Z_STRLEN(events[i]);
+            arg_idx++;
+        }
+        core_args->all_args = all;
+    }
+    core_args->arg_count = arg_idx;
```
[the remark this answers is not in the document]
prateek-kumar-improving: Updated.

valkey_glide_commands.c
nderraugh: I see this is passing in CI. This review might have been on a previous commit.  I'll take another look in a minute to see if this is still a problem.

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,115 @@ public function testMemoryStatsWithRoute()
...
+    }
+
+    public function testLatencyHistory()
+    {
+        $this->triggerLatencySpike();
+
+        // Use default route (allPrimaries with Special policy) to ensure we hit the node with the spike
+        $result = $this->valkey_glide->latencyHistory('command');
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
```
nderraugh, line 923: Can we check the shape of the data is correct here, rather than just the count?

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,115 @@ public function testMemoryStatsWithRoute()
...
+        $this->assertEquals(0, count($unknown));
+    }
+
+    public function testLatencyLatest()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyLatest();
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
```
nderraugh, line 952: Same issue again: shape, not just count.

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,115 @@ public function testMemoryStatsWithRoute()
...
+        $this->triggerLatencySpike();
+
+        // Assert reset count > 0 (aggregated across all primaries)
+        $result = $this->valkey_glide->latencyReset();
+        $this->assertIsInt($result);
+        $this->assertGT(0, $result);
+
+        // History should be empty after reset on any node
+        $history = $this->valkey_glide->latencyHistory('command', 'randomNode');
+        $this->assertEquals(0, count($history));
```
nderraugh: Same issue again: shape, not just count.

valkey_glide_commands.c
nderraugh: Yeah, I think this is still valid.

But the CI is passing now because of the assertions being a bit under-constrained.   I added comments where we should dig into the responses in our assertions.

valkey_glide_cluster.stub.php
xShinnRyuu: `latencyReset(mixed $route = null, string ...$events)` with route as an optional first parameter. This is valid PHP and would let one method serve both cases.

xShinnRyuu: discussed offline. The current implementation is fine.

xShinnRyuu (approved):

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/bb872ff06710a7109c3a2819630b35e232483501

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/bb872ff06710a7109c3a2819630b35e232483501

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/bb872ff06710a7109c3a2819630b35e232483501

valkey_glide_commands.c
prateek-kumar-improving: Sounds good. Checking the comments below.

yipin-chen (approved):

valkey_glide_cluster.stub.php
```diff
@@ -1035,6 +1035,36 @@ public function memoryPurge(mixed $route = null): ValkeyGlideCluster|bool|array|
      */
     public function memoryStats(mixed $route = null): ValkeyGlideCluster|array|false;
 
+    /**
+     * @see ValkeyGlide::latencyHistory
+     *
+     * @param mixed $route Optional routing: 'allPrimaries', 'allNodes', 'randomNode', or a specific node address.
+     */
+    public function latencyHistory(string $event, mixed $route = null): ValkeyGlideCluster|array|false;
```
nderraugh, line 1050: **Explicit null routes fail despite being documented defaults**

 The cluster signatures allow:
```php
latencyHistory($event, null)
latencyLatest(null)
```
(valkey_glide_cluster.stub.php:1043,1050). However, any supplied argument sets has_route, and the route parser accepts only strings or arrays. Explicit null therefore fails instead of behaving like an omitted route. 

The public declarations explicitly use null as the optional default:

valkey_glide_cluster.stub.php:1043
```php
public function latencyHistory(
    string $event,
    mixed $route = null
): ValkeyGlideCluster|array|false;
```
valkey_glide_cluster.stub.php:1050
```php

public function latencyLatest(
    mixed $route = null
): ValkeyGlideCluster|array|false;
```
But valkey_glide_commands.c treats any supplied second argument as a route, including null:

```c
if (args_count == 1) {
    core_args.has_route   = 1;
    core_args.route_param = &args[0];
}
```
That occurs in both execute_latency_history_command() and execute_latency_latest_command().

The resulting path is:
1. execute_core_command() sees has_route == 1.
1. It calls execute_command_with_route().
1. parse_cluster_route() accepts only strings and arrays.
1. null fails parsing.
1. No command is sent, and the PHP method returns false.
Omitting the argument leaves has_route == 0 and correctly applies Core’s default route.

**Why CI passes**
There are no tests for:
```php
latencyHistory('command', null)
latencyLatest(null)
```
Existing tests cover omitted routes and valid randomNode routes only.

Additional tests you could add:
```php
    public function testLatencyHistoryWithNullRoute()
    {
        $result = $this->valkey_glide->latencyHistory('command', null);

        // Explicit null should behave like an omitted route, not return false.
        $this->assertIsArray($result);
    }

    public function testLatencyLatestWithNullRoute()
    {
        $result = $this->valkey_glide->latencyLatest(null);

        // Explicit null should behave like an omitted route, not return false.
        $this->assertIsArray($result);
    }
```

prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/9bea693f73125419aefb3f98f15a0925a43c93f9

valkey_glide_commands.c
prateek-kumar-improving: Updated

- **Line 683**: latencyHistory — changed from process_core_array_result to process_cluster_array_result
- **Line 726**: latencyLatest — changed from process_core_array_result to process_cluster_array_result

The new processor is defined at lines 622–629:

```
static int process_cluster_array_result(CommandResponse* response,
                                        void*            output,
                                        zval*            return_value) {
    if (!response || !return_value) {
        return 0;
    }
    return command_response_to_zval(
        response, return_value, COMMAND_RESPONSE_ASSOSIATIVE_ARRAY_MAP, true);
}
```

tests/ValkeyGlideClusterTest.php
```diff
@@ -894,6 +894,154 @@ public function testMemoryStatsWithRoute()
...
+        }
+        $this->assertTrue($hasEntries);
+    }
+
+    public function testLatencyHistoryWithRoute()
+    {
+        $this->triggerLatencySpike();
+
+        $result = $this->valkey_glide->latencyHistory('command', 'randomNode');
+        $this->assertIsArray($result);
```
nderraugh, line 1027: For the next three tests, the tests could be stronger.

They assert only array, array, and int. They would pass with empty history or a zero reset count, so they do not prove that the routed command affected the intended node.

If we think about the shape of the response you are expecting, can we test for something specific that will establish correctness?

valkey_glide_cluster.stub.php
```diff
@@ -1035,6 +1035,36 @@ public function memoryPurge(mixed $route = null): ValkeyGlideCluster|bool|array|
      */
     public function memoryStats(mixed $route = null): ValkeyGlideCluster|array|false;
 
+    /**
+     * @see ValkeyGlide::latencyHistory
+     *
+     * @param mixed $route Optional routing: 'allPrimaries', 'allNodes', 'randomNode', or a specific node address.
+     */
```
nderraugh, line 1049:
```suggestion
     *                     If omitted, the command is routed to all primary nodes.
     *
     * @return ValkeyGlideCluster|array|false For a single-node route, returns a list of
     *                                        [timestamp, duration] entries. For 'allPrimaries',
     *                                        'allNodes', or the default route, returns an associative
     *                                        array that maps each node address to its list of entries.
     *                                        Returns false on failure.
     */
```

valkey_glide_cluster.stub.php
```diff
@@ -1035,6 +1035,36 @@ public function memoryPurge(mixed $route = null): ValkeyGlideCluster|bool|array|
...
+    public function latencyHistory(string $event, mixed $route = null): ValkeyGlideCluster|array|false;
+
+    /**
+     * @see ValkeyGlide::latencyLatest
+     *
+     * @param mixed $route Optional routing: 'allPrimaries', 'allNodes', 'randomNode', or a specific node address.
+     */
```
nderraugh, line 1063:
```suggestion
     *                     If omitted, the command is routed to all primary nodes.
     *
     * @return ValkeyGlideCluster|array|false For a single-node route, returns a list of
     *                                        [event, timestamp, latest latency, maximum latency]
     *                                        entries. For 'allPrimaries', 'allNodes', or the default
     *                                        route, returns an associative array that maps each node
     *                                        address to its list of entries. Returns false on failure.
     */
```

valkey_glide_cluster.stub.php
```diff
@@ -1035,6 +1035,36 @@ public function memoryPurge(mixed $route = null): ValkeyGlideCluster|bool|array|
...
+    public function latencyLatest(mixed $route = null): ValkeyGlideCluster|array|false;
+
+    /**
```
nderraugh, line 1066:
```suggestion
    /**
     * Resets latency data on all primary nodes.
     *
     * @param string ...$events The latency events to reset. If omitted, all latency events are reset.
     *
     * @return ValkeyGlideCluster|int|false The sum of the numbers of events reset on all primary nodes.
     *                                      Returns false on failure.
     *
```

valkey_glide_cluster.stub.php
```diff
@@ -1035,6 +1035,36 @@ public function memoryPurge(mixed $route = null): ValkeyGlideCluster|bool|array|
...
+     * @see ValkeyGlide::latencyReset
+     */
+    public function latencyReset(string ...$events): ValkeyGlideCluster|int|false;
+
+    /**
+     * Resets latency data with explicit routing.
+     *
+     * @param mixed  $route  Routing: 'allPrimaries', 'allNodes', 'randomNode', or a specific node address.
+     *
+     * @return ValkeyGlideCluster|int|false Returns the number of events reset. Returns false on failure.
```
nderraugh:
```suggestion
     * @param string ...$events The latency events to reset. If omitted, all latency events are reset.
     *
     * @return ValkeyGlideCluster|int|false Returns the number of events reset. Returns false on failure.
     * @return ValkeyGlideCluster|int|false The number of events reset. When the route targets multiple
     *                                      nodes, returns the sum of the per-node counts. Returns false
     *                                      on failure.
```

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: We have already tested here: testLatencyLatest(). Don't think so it is required but I can add here as well.

But in general we don't need to test the Valkey server's correctness, we need to test the client. That will be tested by Valkey itself. As long as the response is returned and is as expected from the server and checked in one of tests, it should be okay.

nderraugh: That's a good point. I lost track of that.

nderraugh (approved):

valkey_glide_cluster.stub.php
prateek-kumar-improving: `* @param string ...$events The latency events to reset. If omitted, all latency events are reset.`
This was removed because it caused a build error. doc does not allow `...`

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/a0c1a6a406818fb50487cb106c7553cda511b483

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit: https://github.com/valkey-io/valkey-glide-php/pull/283/changes/a0c1a6a406818fb50487cb106c7553cda511b483

valkey_glide_cluster.stub.php
prateek-kumar-improving: The PHP arginfo generator (gen_stub.php) parses `@param` tags to generate the *_arginfo.h headers needed for compilation. It expects the format: `@param type $variable_name description`
`@param string ...$events` has the `...` prefix on the variable which the parser can't handle. It throws:

ValkeyGlide::latencyReset(): @param doesn't contain a variable name or has an invalid format "string ...$events..."


Then arginfo generation fails → valkey_glide_cluster_arginfo.h is never created → compilation fails with fatal error: valkey_glide_cluster_arginfo.h: No such file or directory.

**This is a limitation of the stub-to-arginfo tooling (build/gen_stub.php), not a code issue. Every other variadic method in the project (bitop, del, blPop, info, etc.) also omits the @param for its variadic parameter for the same reason.**

PR #292, opened 2026-08-07
feat: Add lib_name and client_info_tag to client configuration

Jonathan-Improving:
## Summary

Adds two new parameters to `ValkeyGlide::connect()` and `ValkeyGlideCluster::__construct()`:

1. **`lib_name`** — overrides the `CLIENT SETINFO LIB-NAME` value. When unset, Rust core default (`GlidePHP`) is preserved.
2. **`client_info_tag`** — appends a parenthesized tag to the resolved library name, e.g. `GlidePHP(my-framework:1.0)`. Follows the ioredis convention for framework attribution.

Closes tracking for PHP parity with valkey-io/valkey-glide#6389 (Python implementation).

## Behavior Matrix

| Config | Resulting `lib-name` |
|---|---|
| _(none)_ | `GlidePHP` (Rust core default) |
| `client_info_tag='fw:1.0'` | `GlidePHP(fw:1.0)` |
| `lib_name='custom'` | `custom` |
| `lib_name='custom'`, `client_info_tag='fw:1.0'` | `custom(fw:1.0)` |

## What Changed

- **`common.h`** — Added `lib_name` and `client_info_tag` fields to both the base client configuration struct and the common constructor params struct.
- **`valkey_glide.c`** — Parse new params in `connect()`, validate `client_info_tag` has no whitespace, pass to connection builder.
- **`valkey_glide_cluster.c`** — Same for cluster client.
- **`valkey_glide_core_commands.c`** — Compose `lib_name` from override and/or tag, set `conn_req.lib_name` (protobuf field 19).
- **`valkey_glide.stub.php`** / **`valkey_glide_cluster.stub.php`** — PHP-visible parameter documentation.
- **`src/client_constructor_mock.c`** / **`.stub.php`** — Mock support for unit testing.
- **`tests/ConnectionRequestTest.php`** — Unit tests for protobuf serialization.
- **`tests/ValkeyGlideFeaturesTest.php`** — Integration tests verifying CLIENT INFO output.

## Testing

- Unit tests via `ConnectionRequestTest` (protobuf field assertions, whitespace validation)
- Integration tests via `ValkeyGlideFeaturesTest` (real connection CLIENT INFO verification)

## Notes

- `lib_name` protobuf field 19 already exists in ConnectionRequest and is consumed by Rust core (Java and Python clients already set it).
- No proto or Rust-core changes required — this is a PHP binding-only change.

### DIFF
--- a/src/client_constructor_mock.stub.php
+++ b/src/client_constructor_mock.stub.php
@@ -105,6 +105,16 @@ class ClientConstructorMock
      *                                           ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD,
      *                                           ValkeyGlide::NODE_DISCOVERY_MODE_STATIC, or
      *                                           ValkeyGlide::NODE_DISCOVERY_MODE_DISCOVER_ALL.
+     * @param callable|null $address_resolver    Custom address resolver callback.
+     * @param string|null $lib_name              Override the library name sent via CLIENT SETINFO LIB-NAME (default: "GlidePHP").
+     *                                           Empty is treated as UNSET. A non-empty value must contain only
+     *                                           printable ASCII, excluding space, '(' and ')'
+     *                                           (^[\x21-\x27\x2A-\x7E]+$). Parentheses are reserved for the
+     *                                           binding's own "(tag)" composition.
+     * @param string|null $client_info_tag       Tag appended to the resolved library name as "<resolved-lib-name>(tag)" for
+     *                                           framework attribution. Empty is treated as UNSET, so no "(tag)"
+     *                                           suffix is added. A non-empty value must contain only printable
+     *                                           ASCII, excluding space, '(' and ')' (^[\x21-\x27\x2A-\x7E]+$).
      */
     public static function simulate_standalone_constructor(
         ?array $addresses = null,
@@ -124,6 +134,8 @@ class ClientConstructorMock
         ?callable $address_resolver = null,
         ?array $circuit_breaker = null,
         int $node_discovery_mode = ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD,
+        ?string $lib_name = null,
+        ?string $client_info_tag = null,
     ): \Connection_request\ConnectionRequest;
 
     /**
@@ -152,6 +164,17 @@ class ClientConstructorMock
      *                                                'compression_level' => 3, 'min_compression_size' => 256].
      * @param array|null $client_side_cache           Client-side cache configuration ['cache_id' => string, 'max_cache_kb' => int,
      *                                                'entry_ttl_ms' => int, 'eviction_policy' => ?int, 'enable_metrics' => bool].
+     * @param callable|null $address_resolver         Custom address resolver callback.
+     * @param string|null $lib_name                   Override the library name sent via CLIENT SETINFO LIB-NAME (default: "GlidePHP").
+     *                                                Empty is treated as UNSET. A non-empty value must contain
+     *                                                only printable ASCII, excluding space, '(' and ')'
+     *                                                (^[\x21-\x27\x2A-\x7E]+$). Parentheses are reserved for
+     *                                                the binding's own "(tag)" composition.
+     * @param string|null $client_info_tag            Tag appended to the resolved library name as "<resolved-lib-name>(tag)" for
+     *                                                framework attribution. Empty is treated as UNSET, so no
+     *                                                "(tag)" suffix is added. A non-empty value must contain only
+     *                                                printable ASCII, excluding space, '(' and ')'
+     *                                                (^[\x21-\x27\x2A-\x7E]+$).
      */
     public static function simulate_cluster_constructor(
         ?array $addresses = null,
@@ -171,5 +194,7 @@ class ClientConstructorMock
         ?array $client_side_cache = null,
         ?callable $address_resolver = null,
         ?array $circuit_breaker = null,
+        ?string $lib_name = null,
+        ?string $client_info_tag = null,
     ): \Connection_request\ConnectionRequest;
 }
--- a/tests/ConnectionRequestTest.php
+++ b/tests/ConnectionRequestTest.php
@@ -1180,6 +1180,233 @@ class ConnectionRequestTest extends \TestSuite
         $this->assertFalse($compression_config->getEnabled());
     }
 
+    // ---- lib_name and client_info_tag tests ----
+
+    public function testStandaloneLibNameOverride()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(lib_name: 'custom-lib');
+        $this->assertEquals('custom-lib', $request->getLibName());
+    }
+
+    public function testClusterLibNameOverride()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor(lib_name: 'custom-lib');
+        $this->assertEquals('custom-lib', $request->getLibName());
+    }
+
+    public function testStandaloneClientInfoTag()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(client_info_tag: 'my-framework:1.0');
+        $this->assertEquals('GlidePHP(my-framework:1.0)', $request->getLibName());
+    }
+
+    public function testClusterClientInfoTag()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor(client_info_tag: 'my-framework:1.0');
+        $this->assertEquals('GlidePHP(my-framework:1.0)', $request->getLibName());
+    }
+
+    public function testStandaloneLibNameWithTag()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(lib_name: 'custom', client_info_tag: 'tag:2.0');
+        $this->assertEquals('custom(tag:2.0)', $request->getLibName());
+    }
+
+    public function testClusterLibNameWithTag()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor(lib_name: 'custom', client_info_tag: 'tag:2.0');
+        $this->assertEquals('custom(tag:2.0)', $request->getLibName());
+    }
+
+    public function testStandaloneLibNameDefault()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor();
+        // When neither is set, the binding sends its own default identity.
+        $this->assertEquals('GlidePHP', $request->getLibName());
+    }
+
+    public function testClusterLibNameDefault()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor();
+        // When neither is set, the binding sends its own default identity.
+        $this->assertEquals('GlidePHP', $request->getLibName());
+    }
+
+    /**
+     * An empty value is treated as absent rather than an error, matching
+     * glide-core's validate_effective_lib_name(). An empty client_info_tag must
+     * therefore produce no "()" suffix, and an empty lib_name must fall back to
+     * the binding default.
+     */
+    public function testStandaloneEmptyClientInfoTagTreatedAsAbsent()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(client_info_tag: '');
+        $this->assertEquals('GlidePHP', $request->getLibName());
+    }
+
+    public function testClusterEmptyClientInfoTagTreatedAsAbsent()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor(client_info_tag: '');
+        $this->assertEquals('GlidePHP', $request->getLibName());
+    }
+
+    public function testStandaloneEmptyLibNameTreatedAsAbsent()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(lib_name: '');
+        $this->assertEquals('GlidePHP', $request->getLibName());
+    }
+
+    public function testClusterEmptyLibNameTreatedAsAbsent()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor(lib_name: '');
+        $this->assertEquals('GlidePHP', $request->getLibName());
+    }
+
+    public function testStandaloneEmptyLibNameWithTagUsesDefaultBase()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(lib_name: '', client_info_tag: 'tag:2.0');
+        $this->assertEquals('GlidePHP(tag:2.0)', $request->getLibName());
+    }
+
+    /** L-9: cluster twin for the empty-lib_name-plus-tag composition. */
+    public function testClusterEmptyLibNameWithTagUsesDefaultBase()
+    {
+        $request = ClientConstructorMock::simulate_cluster_constructor(lib_name: '', client_info_tag: 'tag:2.0');
+        $this->assertEquals('GlidePHP(tag:2.0)', $request->getLibName());
+    }
+
+    /**
+     * M-3: guards the C validator's charset against glide-core's LIB_NAME_PATTERN
+     * character class [\x21-\x27\x2A-\x7E] (glide-core/src/client/mod.rs).
+     *
+     * The entire safety argument for keeping the binding's per-field validator is
+     * C-accept subset-of core-accept. That invariant spans a submodule boundary
+     * bumped independently of this code, so prose comments alone cannot hold it.
+     * If a core bump changes the grammar, this fails loudly here instead of
+     * silently deferring rejection to connect time.
+     *
+     * 0x00 is excluded: a NUL cannot survive PHP -> C marshalling as data.
+     */
+    public function testMetadataCharsetParityWithGlideCore()
+    {
+        foreach (['lib_name', 'client_info_tag'] as $field) {
+            for ($b = 0x01; $b <= 0xFF; $b++) {
+                $ch          = chr($b);
+                $coreAccepts = ($b >= 0x21 && $b <= 0x27) || ($b >= 0x2A && $b <= 0x7E);
+
+                $rejected = false;
+                try {
+                    if ($field === 'lib_name') {
+                        ClientConstructorMock::simulate_standalone_constructor(lib_name: $ch);
+                    } else {
+                        ClientConstructorMock::simulate_standalone_constructor(client_info_tag: $ch);
+                    }
+                } catch (ValkeyGlideException $e) {
+                    $rejected = true;
+                }
+
+                $this->assertTrue(
+                    $rejected === !$coreAccepts,
+                    sprintf(
+                        '%s byte 0x%02X: binding %s but core %s',
+                        $field,
+                        $b,
+                        $rejected ? 'rejects' : 'accepts',
+                        $coreAccepts ? 'accepts' : 'rejects'
+                    )
+                );
+            }
+        }
+    }
+
+    /**
+     * Values that must be rejected for lib_name / client_info_tag.
+     *
+     * Note: the empty string is deliberately NOT listed here. An empty value is
+     * treated as absent (matching glide-core's validate_effective_lib_name),
+     * so it yields the default library name with no "(tag)" suffix rather than
+     * an error. See testStandaloneEmptyClientInfoTagTreatedAsAbsent.
+     *
+     * Parentheses are rejected per field because the binding introduces the only
+     * permitted "(tag)" pair when composing the effective name; a field carrying
+     * its own parenthesis would compose into a name core rejects.
+     */
+    private function invalidMetadataValues(): array
+    {
+        return [
+            'space'          => ['has space'],
+            'tab'            => ["has\ttab"],
+            'newline'        => ["has\nnewline"],
+            'embedded_nul'   => ["framework\0suffix"],
+            'control_bel'    => ["framework\x07"],
+            'del'            => ["framework\x7f"],
+            'non_ascii_utf8' => ['framework-é'],
+            'open_paren'     => ['framework('],
+            'close_paren'    => ['framework)'],
+            'wrapped_parens' => ['(framework)'],
+        ];
+    }
+
+    private function assertMetadataRejected(callable $construct, string $field, string $case = ''): void
+    {
+        $label = $case !== '' ? " (case '{$case}')" : '';
+        try {
+            $construct();
+            // M-5: name the offending case. Without this the leaked-invalid-value
+            // branch reported a bare "(false) !== (true)" with no case identity.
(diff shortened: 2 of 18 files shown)


tests/ConnectionRequestTest.php
```diff
@@ -730,6 +730,71 @@ public function testCompressionClusterDisabled()
...
+
+    public function testClusterClientInfoTagWhitespaceRejected()
+    {
+        try {
+            ClientConstructorMock::simulate_cluster_constructor(client_info_tag: "has\ttab");
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains('client_info_tag must not contain whitespace', $e->getMessage());
+        }
+    }
```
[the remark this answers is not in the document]
nderraugh: One step further than coderabbitai suggests:

**Tag validation is too permissive and NUL values are silently truncated**
`common.h:13-25` rejects only bytes recognized by `isspace()`. It therefore accepts forbidden values such as:
* DEL: "tag\x7f"
* Non-whitespace controls such as BEL
* Printable Unicode such as "framework-é"
* Embedded NUL

The embedded-NUL case also causes silent corruption. PHP provides explicit string lengths, but valkey_glide.c:206-207 drops them when building the base configuration, and valkey_glide_core_commands.c:191-214 uses strlen/snprintf. For example, "framework\0suffix" becomes only "framework".

Required fix: validate every input byte as printable ASCII `^[!-~]+$` i.e. `byte >= 0x21 && byte <= 0x7e`. This rejects NUL and all other forbidden bytes before composition.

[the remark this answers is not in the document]
nderraugh: **lib_name is not validated before connecting**
We need both `lib_name` and `client_info_tag` to independently match printable ASCII ^[!-~]+$.

Make sure to validate both.

PHP provides explicit string lengths, but `valkey_glide.c:206-207` drops them when building the base configuration, and `valkey_glide_core_commands.c:191-214` uses `strlen/snprintf`. For example, "framework\0suffix" becomes only "framework".

Required fix: use one length-aware printable-ASCII validator for both fields in standalone, cluster, and mock constructors.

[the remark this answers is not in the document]
nderraugh: (optional) **no binding-local pure resolver** The other clients (Java, Node, Python, Go) use a pure resolver over raw override and tag values so the full behavior matrix can be tested without a server or FFI.

Without options, that resolver should return the existing PHP default.

Instead, composition is embedded directly inside `create_connection_request()` at `valkey_glide_core_commands.c:191-214`, and the no-options branch `sets conn_req.lib_name = NULL`. The corresponding test explicitly expects an empty request field at `ConnectionRequestTest.php:772-776`.

The server currently still sees `GlidePHP` because the compile-time core fallback supplies it, but PHP request behavior now depends on build configuration. A build without the expected GLIDE_NAME would report UnknownClient, contrary to preserving PHP’s default identity.

Required fix: create a pure resolver returning `GlidePHP`, override, `GlidePHP(tag)`, or `override(tag)`.

valkey_glide_core_commands.c
```diff
@@ -188,6 +188,31 @@ uint8_t* create_connection_request(size_t*                                   len
...
+        conn_req.lib_name = composed_lib_name;
+    } else if (config->lib_name && strlen(config->lib_name) > 0) {
+        /* Full override, no tag */
+        conn_req.lib_name = config->lib_name;
+    } else if (config->client_info_tag && strlen(config->client_info_tag) > 0) {
+        /* Default lib name (GlidePHP) + (tag) — Rust core sets default,
+           but if tag is provided we must compose here with the known default */
+        size_t composed_len = strlen("GlidePHP") + 1 + strlen(config->client_info_tag) + 1 + 1;
+        composed_lib_name   = emalloc(composed_len);
+        snprintf(composed_lib_name, composed_len, "GlidePHP(%s)", config->client_info_tag);
```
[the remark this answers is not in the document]
Jonathan-Improving: Mitigated. Both `lib_name` and `client_info_tag` are now validated as non-empty printable ASCII (`^[!-~]+$`, i.e. every byte in 0x21–0x7e) at all four constructor entry points (standalone, cluster, and both constructor mocks) *before* any request composition, via the shared `client_metadata_validation_error()` helper in `common.h`. This rejects embedded NUL (as well as control bytes, DEL, whitespace, non-ASCII, and empty strings), so a value like `"custom\0suffix"` is refused up front rather than being silently truncated by the later `strlen`/`snprintf`.

Composition itself was also extracted into a single pure resolver `valkey_glide_resolve_lib_name()` (`valkey_glide_core_commands.c`) with the default name sourced from one `VALKEY_GLIDE_LIB_NAME` constant, removing the previously hardcoded `"GlidePHP"` in the tag-only branch.

Commits: 4a7e5df (printable-ASCII validation across all entry points) and b9becdc (single resolver + shared validation helper). Added table-driven rejection tests covering "", "\n", NUL, BEL (0x07), DEL (0x7f), and non-ASCII for both fields via the standalone and cluster mock constructors.

valkey_glide_core_commands.c
```diff
@@ -43,6 +43,38 @@ extern zend_class_entry* get_valkey_glide_exception_ce();
...
+    if (have_tag) {
+        size_t composed_len = strlen(base) + strlen(client_info_tag) + 3; /* '(' ')' '\0' */
+        char*  composed     = emalloc(composed_len);
+        snprintf(composed, composed_len, "%s(%s)", base, client_info_tag);
+        return composed;
+    }
+    if (have_lib) {
+        return estrdup(lib_name);
+    }
+    return NULL;
```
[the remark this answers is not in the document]
Jonathan-Improving: Done in ad6daed. `valkey_glide_resolve_lib_name()` now returns `estrdup(VALKEY_GLIDE_LIB_NAME)` (i.e. `GlidePHP`) in the both-unset branch instead of `NULL`, so the binding always sends its own default identity rather than depending on the Rust-core fallback / build configuration. Updated both `testStandaloneLibNameDefault` and `testClusterLibNameDefault` to expect `GlidePHP`. Build is clean (-Werror) and ConnectionRequestTest passes 72/72.

valkey_glide_core_commands.c
```diff
@@ -188,6 +219,11 @@ uint8_t* create_connection_request(size_t*                                   len
     /* Set client name */
     conn_req.client_name = config->client_name ? config->client_name : NULL;
 
+    /* Set lib_name via the pure resolver (single source of truth for the
+     * default identity; NULL means "let the Rust core apply its default"). */
+    char* composed_lib_name = valkey_glide_resolve_lib_name(config->lib_name, config->client_info_tag);
+    conn_req.lib_name       = composed_lib_name;
```
prateek-kumar-improving: **Comment no longer matches behavior — `lib_name` is never NULL now.**

`valkey_glide_resolve_lib_name()` returns `estrdup(VALKEY_GLIDE_LIB_NAME)` ("GlidePHP") when nothing is set, so `composed_lib_name` / `conn_req.lib_name` is always non-NULL. The "NULL means let the Rust core apply its default" clause is now misleading — the binding always sends its own identity and never defers to the core fallback.

 Please update the comment to reflect that the resolver always returns a non-NULL string, e.g.:

```c
/* Resolve lib_name via the pure resolver. Always returns a heap-owned,
 * non-NULL string (defaults to VALKEY_GLIDE_LIB_NAME), so the binding
 * always reports its own identity rather than relying on the core default.
 * Freed after packing below. */
```

Is the glide-core default of any use?

Jonathan-Improving: Fixed in 8ec4689 — reworded the call-site comment to state the resolver always returns a heap-owned, non-NULL string (defaults to `VALKEY_GLIDE_LIB_NAME`) and that the binding always reports its own identity rather than deferring to the core default.

Re "is the glide-core default of any use?": for this PHP binding, no — since `valkey_glide_resolve_lib_name()` now always sends a concrete `lib_name`, the Rust-core `GLIDE_NAME` fallback is never exercised on the connection-request path. Sending our own default was the intent of the earlier review (avoid depending on core build configuration for the reported identity), so the core default is effectively dead for us here; I left the core untouched since it is shared across language bindings.

prateek-kumar-improving (approved):

prateek-kumar-improving:
@Jonathan-Improving 
Started the workflows and lint is failing: https://github.com/valkey-io/valkey-glide-php/actions/runs/33202363282/job/98992805530?pr=292

valkey_glide.stub.php
```diff
@@ -433,6 +434,11 @@ public function __construct()
      *                                      NODE_DISCOVERY_MODE_STANDARD (default),
      *                                      NODE_DISCOVERY_MODE_STATIC, or
      *                                      NODE_DISCOVERY_MODE_DISCOVER_ALL
+     * @param string|null $lib_name Override the library name sent via CLIENT SETINFO LIB-NAME (default: "GlidePHP").
+     *      Must contain only printable ASCII characters (^[!-~]+$).
```
nderraugh: That expression: `(^[!-~]+$)`

permits ( and ), although the implementation rejects them;
requires at least one character, although empty strings are treated as unset.
Document the actual contract: empty strings are treated as unset; non-empty values must contain U+0021–U+007E except ( and ).

valkey_glide_cluster.stub.php
```diff
@@ -272,10 +272,17 @@ class ValkeyGlideCluster
      * @param array|null $client_side_cache     Client-side cache configuration array from ClientSideCache::toArray():
      *                                          ['cache_id' => string, 'max_cache_kb' => int, 'entry_ttl_ms' => int,
      *                                          'eviction_policy' => ?int, 'enable_metrics' => bool]
+     * @param callable|null $address_resolver   Custom address resolver callback
      * @param array|null $circuit_breaker       Circuit breaker configuration:
      *                                          ['window_size_ms' => int, 'failure_rate_threshold' => float,
      *                                           'min_errors' => int, 'open_timeout_ms' => int,
      *                                           'count_timeouts' => bool, 'consecutive_successes' => int]
+     * @param string|null $lib_name             Override the library name sent via CLIENT SETINFO LIB-NAME (default: "GlidePHP").
+     *                                          Must contain only printable ASCII characters (^[!-~]+$).
```
nderraugh: Same as https://github.com/valkey-io/valkey-glide-php/pull/292/changes#r3907550729

src/client_constructor_mock.stub.php
```diff
@@ -105,6 +105,11 @@ class ClientConstructorMock
      *                                           ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD,
      *                                           ValkeyGlide::NODE_DISCOVERY_MODE_STATIC, or
      *                                           ValkeyGlide::NODE_DISCOVERY_MODE_DISCOVER_ALL.
+     * @param callable|null $address_resolver    Custom address resolver callback.
+     * @param string|null $lib_name              Override the library name sent via CLIENT SETINFO LIB-NAME (default: "GlidePHP").
+     *                                           Must contain only printable ASCII characters (^[!-~]+$).
```
nderraugh: Same comment about `(^[!-~]+$)`.

valkey_glide.c
```diff
@@ -1280,6 +1300,16 @@ PHP_METHOD(ValkeyGlide, connect) {
         RETURN_FALSE;
     }
 
+    /* Validate lib_name and client_info_tag (printable ASCII, non-empty) */
```
nderraugh: Missing similar feature for `valkey_glide_monitor.c`.  We need to cover those monitor clients too.

nderraugh: You can compare to Go in valkey-glide: 

NewMonitorClient receives a normal ClientConfiguration and calls:

```go
request, err := cfg.ToProtobuf()
```
That shared serializer:
* resolves lib_name and client_info_tag in resolveLibName;
* assigns the resolved value to request.LibName in baseClientConfiguration.toProtobuf;
* exposes the options through WithLibName and WithClientInfoTag.

So ordinary and monitor clients use the same configuration serialization and produce the same effective name.

valkey_glide_pubsub_common.c
```diff
@@ -63,12 +63,14 @@ void cond_wait(cond_t* c, mutex_t* m) {
...
-#define NSEC_PER_MSEC 1000000LL
-#endif
+/* M-6: project-owned names. Previously these were NSEC_PER_SEC / NSEC_PER_MSEC
+ * behind #ifndef guards, which collided with macOS's mach/clock_types.h. The
+ * guards resolved the -Werror build failure but let the platform's definition win
+ * silently -- and the platform spells them `ull` where we mean `LL`, so the
+ * timeout arithmetic below would quietly convert to unsigned. Owning the names
+ * removes both the collision and the signedness ambiguity. */
+#define VALKEY_GLIDE_NSEC_PER_SEC 1000000000LL
+#define VALKEY_GLIDE_NSEC_PER_MSEC 1000000LL
```
Jonathan-Improving: This looks like a platform-dependent failure point.

valkey_glide.stub.php
Jonathan-Improving: MITIGATED in 10d99ec (charset) and 947dbd0 (empty-is-unset). Both points were correct, and the second one nobody else caught.

**Point 1 — `(` and `)`.** `^[!-~]+$` is `0x21-0x7E`, which includes `(` `0x28` and `)` `0x29`, and the implementation rejects both. Corrected to `^[\x21-\x27\x2A-\x7E]+$` across all eight docblock lines and the two C comments that carried the same expression. There are no occurrences of `[!-~]` left anywhere outside the submodule.

**Point 2 — the `+` requiring at least one character.** You are right, and this was a real gap rather than a wording nit: the `+` documented a contract that forbids empty, while the implementation has always treated empty as unset. So even after the charset fix the docs still contradicted the behaviour. The docblocks now state it explicitly — an empty string or null is treated as UNSET (yielding the default, and adding no `(tag)` suffix), and only non-empty values are charset-constrained.

I also documented *why* the parens are excluded rather than just that they are, since the reason is not obvious from the range: the binding itself introduces the single `(tag)` pair when composing `<lib_name>(<client_info_tag>)`, so a field carrying its own parenthesis would compose into a name glide-core rejects. `client_info_tag` is the supported way to get a parenthesised suffix.

For reference, the accepted set is now byte-for-byte identical to glide-core `validate_effective_lib_name()` (`glide-core/src/client/mod.rs`), whose class is `[\x21-\x27\x2A-\x7E]`, with empty treated as absent at both layers. A byte-sweep test (`testMetadataCharsetParityWithGlideCore`, `0x01-0xFF`, both fields) pins the binding against that class so a future core bump cannot silently diverge; I verified it is falsifiable by widening the expected class and watching it fail on exactly `0x28`/`0x29`.

Visible on the current PR head `947dbd0`.

valkey_glide_cluster.stub.php
Jonathan-Improving: MITIGATED — same fix applied here, see the detailed reply on the `valkey_glide.stub.php` thread.

`valkey_glide_cluster.stub.php` now documents the corrected charset `^[\x21-\x27\x2A-\x7E]+$` for both `lib_name` and `client_info_tag`, and states that an empty string or null is treated as UNSET rather than being charset-constrained.

Worth noting there were **five** docblock groups carrying this expression, not three: the two you flagged plus two in `src/client_constructor_mock.stub.php` (standalone and cluster) and, after adding monitor support in response to your other comment, one in `valkey_glide_monitor.stub.php`. All five are corrected, and a repo-wide grep for `[!-~]` outside the submodule returns nothing. A single search-and-replace would have missed some because each copy used a different continuation indent, so they were done individually and verified by grep.

Visible on the current PR head `947dbd0`.

src/client_constructor_mock.stub.php
Jonathan-Improving: MITIGATED — same fix, see the detailed reply on the `valkey_glide.stub.php` thread.

This file carried **two** copies of the expression, not one — the standalone docblock and the cluster docblock. Both are corrected to `^[\x21-\x27\x2A-\x7E]+$` with the empty-is-unset contract stated, and the two matching C comments in `src/client_constructor_mock.c` that described the charset as `^[!-~]+$` were corrected as well.

Visible on the current PR head `947dbd0`.

valkey_glide.c
Jonathan-Improving: MITIGATED in 947dbd0 — `ValkeyGlideMonitor` now accepts and validates both options.

Your Go comparison was the useful part. PHP already had the shape you described — the monitor constructor routes through the same `valkey_glide_build_client_config_base()` and `create_connection_request()` as the standalone and cluster paths, which is the equivalent of Go sharing `cfg.ToProtobuf()` / `resolveLibName` — it simply never accepted the two parameters, so they could never reach that shared serializer. That is now wired through:

- `valkey_glide_monitor.c`: arity `(0,6)` -> `(0,8)`, adding `lib_name` then `client_info_tag` last, with matching stub signature and regenerated arginfo.
- Validation uses the **same** `client_metadata_validation_error()` from `common.h` as the other three entry points, not a copy — so accept/reject behaviour cannot drift between them. The gate sits before `valkey_glide_build_client_config_base()`, so no ungated path reaches request composition.
- Resolution is unchanged and shared, so a monitor client produces the same effective name as an ordinary client, per your point.

Verified server-side rather than by construction, matching the `cmd=monitor` row in `CLIENT LIST`:

| config | observed `lib-name` |
|---|---|
| none | `GlidePHP` |
| `client_info_tag=framework:1.2` | `GlidePHP(framework:1.2)` |
| `lib_name=custom-lib` | `custom-lib` |
| both | `custom-lib(framework:1.2)` |

A space-bearing tag and a parenthesis-bearing `lib_name` each raise `ValkeyGlideException` naming the offending field. Covered by `testMonitorReportsComposedLibName` and `testMonitorRejectsInvalidMetadata`.

One note on heap lifetime, since the monitor previously had these fields NULL-by-memset and so never allocated a composed name: the allocation and both frees live entirely inside `create_connection_request()` (allocated at `valkey_glide_core_commands.c:273`, freed on both the failure and success exits), and the monitor never touches `config->lib_name` after the call. No caller-side free is needed and no new leak surface is introduced.

Visible on the current PR head `947dbd0`.

Jonathan-Improving:
## Status of @nderraugh review comments

All four open threads are mitigated and replied to individually. Summary for reviewer convenience:

| # | Comment | Status | Where |
|---|---|---|---|
| 1 | `^[!-~]+$` permits `(` and `)`, which the implementation rejects | **Mitigated** | `10d99ec` |
| 2 | `^[!-~]+$` requires ≥1 char, contradicting empty-is-unset | **Mitigated** | `947dbd0` |
| 3 | Same on `valkey_glide_cluster.stub.php` / `src/client_constructor_mock.stub.php` | **Mitigated** | `10d99ec`, `947dbd0` |
| 4 | Missing feature for `valkey_glide_monitor.c` | **Mitigated** | `947dbd0` |

Earlier comments from the August round (whitespace-only validation too permissive, embedded NUL truncation, `lib_name` not validated, no binding-local pure resolver) were addressed in `4a7e5df`, `2be1f50`, `b9becdc` and `fd5be87`; that thread is resolved.

### Detail

**Docblock contract (1-3).** The documented charset is now `^[\x21-\x27\x2A-\x7E]+$` — printable ASCII excluding space, `(` and `)` — across all **five** docblock groups (`valkey_glide.stub.php`, `valkey_glide_cluster.stub.php`, two in `src/client_constructor_mock.stub.php`, and `valkey_glide_monitor.stub.php`) plus the two matching C comments. Empty/null is documented as UNSET. No `[!-~]` occurrences remain outside the submodule.

The second point was a genuine gap rather than a wording nit: the `+` documented a contract forbidding empty while the implementation has always treated empty as unset, so the docs contradicted the behaviour even after the charset correction.

**Monitor clients (4).** `ValkeyGlideMonitor` accepts both options, validated by the **same** shared predicate as the other three entry points (four entry points, one validator, so accept/reject cannot drift). Verified server-side via the `cmd=monitor` row in `CLIENT LIST` for all four matrix cases, with invalid values rejected. New tests: `testMonitorReportsComposedLibName`, `testMonitorRejectsInvalidMetadata`.

### Not addressed (deliberate, and not blocking)

- **`ClientConstructorMock` ships unconditionally in release builds** — compiled, registered and packaged with no `PHP_ARG_ENABLE` gate, so a test-only class is public API surface. Pre-existing: `config.m4`, `package.xml` and `Makefile.frag` are untouched by this branch and the mock exists at the merge base. Wants its own PR.
- **`./lint-c.sh` exits non-zero.** Every violation attributable to this branch is fixed and all authored `.c`/`.h` files are now clang-format clean under whole-file checking. The remaining 8485 violations are **100%** in generated `*_arginfo.h` files, which POLICY.md forbids committing, `.gitignore` excludes, and the build regenerates — so the gate cannot pass in a working tree. The script also hard-pins clang-format 18.x. Both look like defects in the gate rather than in this branch.
- **Structural debt this change extends rather than introduces** — the 4× duplicated call-and-throw wrapper, and `create_connection_request()`’s 24 positional parameters. Better as a standalone refactor.

### Verification

PHP 8.3 + Valkey 9.1.1. `ConnectionRequestTest` 107, `ValkeyGlideFeaturesTest` 94, `ValkeyGlideClusterFeaturesTest` 85, `ValkeyGlideMonitorTest` 14 — all passing. Two pre-existing unrelated failures persist and are not caused by this branch: `testFailoverSucceeds` (suite fragility around replication state) and `testGetCommandWithConsoleLogging` (Homebrew PHP disables dynamic `dl()`).

Submodule gitlink is `49f2b6d21`, committed and visible on this PR head.

valkey_glide_pubsub_common.c
nderraugh: I don’t see a platform-dependent failure in this constant:

* LL is standard C.
* long long is guaranteed to be at least 64 bits.
* timeout_ms is an unsigned int; even its maximum value multiplied by 1,000,000 fits comfortably in signed 64-bit arithmetic.
* The calculation normalizes tv_nsec into the required [0, 1_000_000_000) range.
* Windows uses SleepConditionVariableCS and bypasses this arithmetic entirely.
* The project-prefixed macro actually fixes a previous macOS portability problem: generic NSEC_PER_* names collided with SDK definitions and could change the arithmetic to unsigned.

Is there a specific platform or type behavior you’re concerned about?

Jonathan-Improving: No, that clears up my concern, thanks.

nderraugh (approved):

Issue #287, opened 2026-07-29
Remove skipped empty tests from cluster tests if possible and remove duplicate tests


prateek-kumar-improving:
1. Remove skipped empty tests from cluster tests if possible
2. Also, try to remove if there are any duplicate tests.

PR #293, opened 2026-08-10
feat(php): implement CLIENT TRACKINGINFO command

prateek-kumar-improving:
### Summary

Implements `CLIENT TRACKINGINFO` command for both standalone and cluster clients.

`CLIENT CACHING` and `CLIENT TRACKING` are **not** exposed as public methods — they will be enabled through connection-level server-assisted client-side caching configuration (see #296). 

### Issue link

This Pull Request is linked to issue: [Add commands for compatibility with other GLIDE clients](https://github.com/valkey-io/valkey-glide-php/issues/236)
Closes partially #236

| Command | Standalone | Cluster | Description |
|---------|-----------|---------|-------------|
| `clientTrackingInfo(mixed $route = null)` | Yes | Yes (with route) | Returns current tracking state (flags, redirect, prefixes) |

### Implementation

- `valkey_glide_commands_3.c` — Added `ClientTrackingInfo` to `determine_client_command_type()` dispatcher + `command_response_to_zval_wrapper` for TRACKINGINFO associative response
- `valkey_glide_core_common.c` — `ClientTrackingInfo` routed as zero-arg command
- `valkey_glide_commands_common.h` — Function declarations and `STANDARD_METHOD_IMPL` macros
- `valkey_z_php_methods.c` / `valkey_glide_cluster.c` — Method registrations
- `valkey_glide.stub.php` / `valkey_glide_cluster.stub.php` — PHP stubs with docblocks

Key design decisions:
- CLIENT TRACKINGINFO uses `process_cluster_array_result` (associative map) for proper key-value response
- Null route handling follows PR #283 pattern (`Z_TYPE(args[0]) != IS_NULL`)
- CLIENT CACHING and CLIENT TRACKING dispatch code is commented out pending #296 (server-assisted client-side caching)

### Limitations

- `CLIENT CACHING` and `CLIENT TRACKING` are not exposed because they are connection-specific commands and unreliable with the multiplexed connection model. They will be enabled once server-assisted client-side caching is implemented at the connection level (#296).

### Testing

**Standalone:**
- `testClientTrackingInfoDefault` — verifies default response shape (flags=off, redirect=-1, prefixes=[])

**Cluster:**
- `testClientTrackingInfoDefault` — verifies response shape (flags/redirect/prefixes)
- `testClientTrackingInfoWithRoute` — with `randomNode` route
- `testClientTrackingInfoWithNullRoute` — null route falls back to default
- `testClientTrackingInfoWithAllNodesRoute` — per-node map, each node off

### Checklist


valkey_glide_commands.c
```diff
@@ -1897,3 +1897,277 @@ int execute_move_command(zval* object, int argc, zval* return_value, zend_class_
...
+         * OPTOUT + NOLOOP */
+        int potential_args = 1; /* ON/OFF */
+        ZEND_HASH_FOREACH_STR_KEY_VAL(ht, key, val) {
+            if (key) {
+                if (zend_string_equals_literal_ci(key, "redirect")) {
+                    potential_args += 2; /* REDIRECT + id */
+                } else if (zend_string_equals_literal_ci(key, "prefixes")) {
+                    if (Z_TYPE_P(val) == IS_ARRAY) {
+                        potential_args +=
+                            1 + zend_hash_num_elements(Z_ARRVAL_P(val)); /* PREFIX + each prefix */
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: capacity now uses `2 * N` for array prefixes to account for both PREFIX keyword and value per entry.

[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: capacity now uses `2 * N` for array prefixes to account for both PREFIX keyword and value per entry.

valkey_glide_commands.c
```diff
@@ -1897,3 +1897,277 @@ int execute_move_command(zval* object, int argc, zval* return_value, zend_class_
...
+
+                if (Z_TYPE_P(val) == IS_LONG) {
+                    args_arr[total_args].type                = CORE_ARG_TYPE_LONG;
+                    args_arr[total_args].data.long_arg.value = Z_LVAL_P(val);
+                } else {
+                    convert_to_long(val);
+                    args_arr[total_args].type                = CORE_ARG_TYPE_LONG;
+                    args_arr[total_args].data.long_arg.value = Z_LVAL_P(val);
+                }
+                total_args++;
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: replaced `convert_to_long(val)` with `zval_get_long(val)` to avoid mutating the caller's options array.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,262 @@ public function testClientPauseBatch()
...
+        // Disable tracking
+        $this->valkey_glide->clientTracking(false);
+    }
+
+    public function testClientTrackingOn()
+    {
+        if (!$this->minVersionCheck('6.0.0')) {
+            $this->markTestSkipped('CLIENT TRACKING requires 6.0.0+');
+            return;
+        }
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: bumped version gate from 6.0.0 to 6.2.0 for all tests that call `clientTrackingInfo()`.

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +852,65 @@ public function testClientPauseBatch()
...
+        // Validate types
+        $this->assertIsArray($info['flags']);
+        $this->assertTrue(is_int($info['redirect']) || is_long($info['redirect']));
+        $this->assertIsArray($info['prefixes']);
+
+        // Default: off, no redirect
+        $this->assertTrue(in_array('off', $info['flags']) || array_key_exists('off', $info['flags']));
+        $this->assertEquals(-1, $info['redirect']);
+        $this->assertEmpty($info['prefixes']);
+    }
```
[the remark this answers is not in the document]
prateek-kumar-improving: Addressed: `testClientTrackingInfoWithAllNodesRoute` explicitly iterates the per-node map; default-route tests assert the single-node shape directly which matches current routing behavior.

valkey_glide_commands.c
```diff
@@ -1897,3 +1897,277 @@ int execute_move_command(zval* object, int argc, zval* return_value, zend_class_
...
+
+        if (potential_args > CORE_ARGS_FIXED_MAX) {
+            all_args    = (core_arg_t*) ecalloc(potential_args, sizeof(core_arg_t));
+            use_dynamic = 1;
+            max_args    = potential_args;
+            /* Copy first arg */
+            all_args[0].type                  = CORE_ARG_TYPE_STRING;
+            all_args[0].data.string_arg.value = on ? "ON" : "OFF";
+            all_args[0].data.string_arg.len   = on ? 2 : 3;
+        }
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: capacity now uses `2 * N` for array prefixes to correctly reserve slots for both PREFIX keyword and value per entry.

valkey_glide.stub.php
```diff
@@ -2427,6 +2427,55 @@ public function clientPause(int $timeout, ?string $mode = null): ValkeyGlide|boo
...
+     *
+     * @return ValkeyGlide|bool|string  Returns true on success, false on failure.
+     *                                  With OPT_REPLY_LITERAL enabled, returns "OK" on success.
+     *
+     * @see https://valkey.io/commands/client-caching
+     */
+    public function clientCaching(bool $enabled): ValkeyGlide|bool|string;
+
+    /**
+     * Enable or disable server-assisted client-side caching for the current connection.
```
affonsov: we are not exposing client caching/tracking. they are only exposed through connection configuration server assisted. please remove this commands

affonsov (changes requested):

valkey_glide.stub.php
prateek-kumar-improving: Currently we are missing the following:
- **#296** — client_side_cache connection parameter so glide-core enables tracking on all connections at creation time
- **#295** — Cluster tests verifying tracking "on" state (blocked by #296)

Removed `clientCaching()` and `clientTracking()` as dedicated public methods. Only `clientTrackingInfo()` remains exposed — matching C#/Java which only expose the read-only info command. CLIENT TRACKING and CLIENT CACHING are still accessible via the generic `client('TRACKING', 'ON', ...)` path for PHPRedis compatibility.

When connection-level client-side caching config is implemented (#296), glide-core will handle these internally.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,114 @@ public function testClientPauseBatch()
...
+    public function testClientTrackingViaGenericClient()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKING with TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // Enable tracking with BCAST via generic client() method
+        $result = $this->valkey_glide->client('TRACKING', 'ON', 'BCAST', 'PREFIX', 'test:', 'NOLOOP');
+        $this->assertTrue($result !== false);
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: changed to `$this->assertTrue($result)` for stronger assertion.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,114 @@ public function testClientPauseBatch()
...
+
+        // Verify via trackinginfo
+        $info = $this->valkey_glide->clientTrackingInfo();
+        $this->assertIsArray($info);
+        $this->assertTrue(in_array('on', $info['flags']) || array_key_exists('on', $info['flags']));
+        $this->assertTrue(in_array('bcast', $info['flags']) || array_key_exists('bcast', $info['flags']));
+        $this->assertTrue(in_array('noloop', $info['flags']) || array_key_exists('noloop', $info['flags']));
+
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed: standalone test now asserts `test:` is in prefixes; cluster AllNodes test asserts flags and prefixes are arrays and prefixes are empty by default.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+* Add `CLIENT TRACKINGINFO` command for standalone and cluster clients ([#293](https://github.com/valkey-io/valkey-glide-php/pull/293))
```
xShinnRyuu, line 7: The changelog only mentions `CLIENT TRACKINGINFO`, but the dispatch table in `determine_client_command_type` now also accepts `CLIENT CACHING` and `CLIENT TRACKING` through the generic `client()` method. That is a real user-visible addition, since previously those subcommands fell through to `InvalidRequest`. Please extend the entry to list all three, matching the PR title.

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +824,89 @@ public function testClientPauseBatch()
...
+
+    public function testClientTrackingInfoWithNullRoute()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // null route should behave like no route (allNodes)
```
xShinnRyuu: The comment says "null route should behave like no route (allNodes)", but the assertions expect a flat `flags`/`redirect`/`prefixes` shape at the top level, which is the single-node result. If null route really returned an allNodes fanout, the outer array would be keyed by node address. Please update this comment to say "null route falls back to the single-node default (random node)" so it matches the actual behavior and the PR description.

valkey_glide_cluster.stub.php
```diff
@@ -1002,6 +1002,25 @@ public function clientPause(int $timeout, ?string $mode = null): ValkeyGlideClus
...
+     *
+     * @return ValkeyGlideCluster|array|false  Returns an associative array with:
+     *                                         - "flags" (array): Active tracking flags.
+     *                                         - "redirect" (int): Client ID for redirection, or -1.
+     *                                         - "prefixes" (array): Key prefixes being tracked.
+     *                                         Returns false on failure.
+     *
+     * @see https://valkey.io/commands/client-trackinginfo
+     */
+    public function clientTrackingInfo(mixed $route = null): ValkeyGlideCluster|array|false;
```
xShinnRyuu, line 1027: Please document what happens when `$route` is null or omitted (the default). Right now the docblock only covers single-node and multi-node routes, but the C path routes through `execute_command` without a route parameter and falls back to the FFI default, which returns a flat map for one node. Users need to know that before they consume the return value.

valkey_glide.stub.php
```diff
@@ -2427,6 +2427,19 @@ public function clientPause(int $timeout, ?string $mode = null): ValkeyGlide|boo
      */
     public function clientUnpause(): ValkeyGlide|bool|string;
 
+    /**
```
xShinnRyuu, line 2430: Consider tightening the return description to state the invariants the tests already check: `flags` is a list of strings (not a map), and `prefixes` is an array of strings that is empty (not null) when no prefixes are tracked. That saves consumers a `var_dump`. The redirect `-1` case is already documented; good.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,114 @@ public function testClientPauseBatch()
...
+        // CLIENT CACHING YES via generic client() method
+        $result = $this->valkey_glide->client('CACHING', 'YES');
+        $this->assertTrue($result !== false);
+
+        // Disable tracking
+        $this->valkey_glide->client('TRACKING', 'OFF');
+    }
+
+    public function testClientTrackingInfoBatch()
```
xShinnRyuu: Just confirming: the standalone stub intentionally has no `$route` parameter (only cluster does), so there is no null-route test on the standalone side. If that is deliberate, ignore this comment.

xShinnRyuu: This test only asserts the three keys exist. The non-batch default test additionally checks `redirect == -1`, empty `prefixes`, and `'off'` in `flags`. The whole point of the batch test is to prove the batch decode matches the direct decode, so please carry over one of those value assertions here.

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +824,89 @@ public function testClientPauseBatch()
...
+
+    public function testClientTrackingInfoWithAllNodesRoute()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // AllNodes route returns per-node map, each node should be off by default
+        $result = $this->valkey_glide->clientTrackingInfo('allNodes');
```
xShinnRyuu, line 874: Nice: this asserts each node's tracking-info map rather than just the outer shape. That is exactly the coverage the allNodes route needs and is stronger than several of the sibling cluster tests. Worth keeping as a pattern for future per-node fanouts.

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +824,89 @@ public function testClientPauseBatch()
...
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
```
xShinnRyuu, line 876: Note for reviewers: `TestSuite::assertGT($maximum, $value)` passes when `$value > $maximum`, so `assertGT(0, count($result))` really does mean "count > 0". The parameter naming reads backwards versus PHPUnit's `assertGreaterThan`, but it is consistent with the rest of the file.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
...
 * Add `LATENCY HISTORY`, `LATENCY LATEST`, and `LATENCY RESET` commands for standalone and cluster clients ([#283](https://github.com/valkey-io/valkey-glide-php/pull/283))
 * Add `MEMORY DOCTOR`, `MEMORY MALLOC-STATS`, `MEMORY PURGE`, and `MEMORY STATS` commands for standalone and cluster clients ([#285](https://github.com/valkey-io/valkey-glide-php/pull/285))
 * Add `REPLICAOF` and `FAILOVER` commands for standalone client ([#282](https://github.com/valkey-io/valkey-glide-php/pull/282))
```
xShinnRyuu, line 10: Please squash the "Fix tests" / "Fix review comments" fix-ups into the feature commit before merge so the history reads as one change. The branch has 8 commits (1 feat + 7 fix-ups); one squashed commit keeps the log readable.

valkey_glide_commands.c
```diff
@@ -1897,3 +1897,49 @@ int execute_move_command(zval* object, int argc, zval* return_value, zend_class_
...
+        return 0;
+    }
+
+    /* Setup core command arguments */
+    core_command_args_t core_args = {0};
+    core_args.glide_client        = valkey_glide->glide_client;
+    core_args.cmd_type            = ClientTrackingInfo;
+    core_args.is_cluster          = is_cluster;
+
+    if (is_cluster) {
```
xShinnRyuu, line 1923: The `"O*"` + `Z_TYPE(args[0]) != IS_NULL` + explicit `zend_throw_exception` pattern here is identical to `latencyHistory` and `latencyLatest`. That is exactly the kind of uniformity that makes this surface refactorable later. Nice.

valkey_glide_commands_3.c
```diff
@@ -499,6 +505,9 @@ static int command_response_to_zval_wrapper(CommandResponse* response,
     if (command_type == ClientList && response->response_type == String) {
         return parse_client_list_response(
             response->string_value, response->string_value_len, return_value);
+    } else if (command_type == ClientTrackingInfo) {
```
xShinnRyuu, line 504: Good that both entry points, the dedicated `clientTrackingInfo()` and the generic `client('TRACKINGINFO', ...)`, decode through `COMMAND_RESPONSE_ASSOSIATIVE_ARRAY_MAP` with `use_false_if_null = true`. That keeps the two surfaces from drifting in what they return.

valkey_glide_core_common.c
```diff
@@ -197,6 +197,7 @@ int prepare_core_args(core_command_args_t* args,
     switch (args->cmd_type) {
         /* Zero argument operations */
         case BgRewriteAof:
+        case ClientTrackingInfo:
```
xShinnRyuu, line 200: Good call routing `ClientCaching` and `ClientTracking` through `prepare_message_args` and `ClientTrackingInfo` through `prepare_zero_args`. That defers option ordering (BCAST, PREFIX, OPTIN, OPTOUT, NOLOOP) to whatever the caller supplies, which is the right layer to enforce it once someone actually wraps a typed API around this. Keeps this PR small and correct.

valkey_glide_commands_3.c
```diff
@@ -349,6 +349,12 @@ static enum RequestType determine_client_command_type(zval* args, int args_count
             return ClientInfo;
         } else if (strcasecmp(subcmd, "NO-EVICT") == 0) {
             return ClientNoEvict;
+        } else if (strcasecmp(subcmd, "CACHING") == 0) {
```
xShinnRyuu: Question: `CLIENT TRACKING` and `CLIENT CACHING` bind state to a single connection, so a multi-node route is meaningless for them. This PR skips the cluster tests that would have exercised this, which is fine as a triage move, but should we add a small guard here that rejects a multi-node route for those two subcommands, rather than relying on the server or the follow-up issue? Either answer is fine, just want to make it a conscious choice.

valkey_glide_cluster.stub.php
```diff
@@ -1002,6 +1002,25 @@ public function clientPause(int $timeout, ?string $mode = null): ValkeyGlideClus
      */
     public function clientUnpause(): ValkeyGlideCluster|bool|string;
 
+    /**
+     * Returns information about the current connection's server-assisted client-side caching state.
+     *
+     * When a route is provided, the command is sent to the specified node.
+     * When a single-node route returns data for one node, the result is the array directly.
+     * When a multi-node route is used, the result is an associative array keyed by node address.
```
xShinnRyuu: Consider naming the route strings inline: `'randomNode'` and single-node addresses yield the flat map; `'allNodes'` and `'allPrimaries'` yield a per-node map keyed by node address. Right now that is only discoverable from the tests.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,114 @@ public function testClientPauseBatch()
...
+        $this->assertTrue(is_int($info['redirect']) || is_long($info['redirect']));
+        $this->assertIsArray($info['prefixes']);
+
+        // Default state: tracking off, no redirect
+        $this->assertTrue(in_array('off', $info['flags']) || array_key_exists('off', $info['flags']));
+        $this->assertEquals(-1, $info['redirect']);
+        $this->assertEmpty($info['prefixes']);
+    }
+
+    public function testClientTrackingInfoAfterEnable()
```
xShinnRyuu: This test enables tracking via `client('TRACKING', 'ON')` and reads via `clientTrackingInfo()`. That is enough for now. When a dedicated `clientTracking()` method lands in a follow-up, please add a variant that enables via the dedicated method so the two entry points stay in sync.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,114 @@ public function testClientPauseBatch()
...
+        $info = $this->valkey_glide->clientTrackingInfo();
+        $this->assertIsArray($info);
+        $this->assertTrue(in_array('on', $info['flags']) || array_key_exists('on', $info['flags']));
+        $this->assertFalse(in_array('off', $info['flags']) || array_key_exists('off', $info['flags']));
+
+        // Disable tracking
+        $this->valkey_glide->client('TRACKING', 'OFF');
+    }
+
+    public function testClientTrackingViaGenericClient()
```
xShinnRyuu: Since this PR keeps `CLIENT TRACKING` on the raw `client()` path, illegal option combinations (`prefixes` without `bcast`, both `optin` and `optout`, non-bool option value) are the server's problem. When a typed `clientTracking()` wrapper lands, please add negative tests for those combinations, since the wrapper is where client-side validation will live.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,114 @@ public function testClientPauseBatch()
...
+        $this->assertIsArray($info);
+        $this->assertTrue(in_array('on', $info['flags']) || array_key_exists('on', $info['flags']));
+        $this->assertTrue(in_array('bcast', $info['flags']) || array_key_exists('bcast', $info['flags']));
+        $this->assertTrue(in_array('noloop', $info['flags']) || array_key_exists('noloop', $info['flags']));
+
+        // Disable tracking
+        $this->valkey_glide->client('TRACKING', 'OFF');
+    }
+
+    public function testClientCachingViaGenericClient()
```
xShinnRyuu: `testClientPauseWithReplyLiteral` proves the reply-literal path for `clientPause`. The generic `client('TRACKING', 'ON')` / `client('CACHING', 'YES')` calls added by this PR flow through the same wrapper and should return `'OK'` under `withOptReplyLiteralEnabled`. Please add one small test so the wrapper's default-branch decoding is pinned.

CHANGELOG.md
prateek-kumar-improving: Updated in this commit id: https://github.com/valkey-io/valkey-glide-php/commit/94db8deb3ff3ee03aff96e5f69e56956ff1d894c

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated in this commit id: https://github.com/valkey-io/valkey-glide-php/commit/501c7dd24797cdea5289826592242c88ee39f27e

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit id: https://github.com/valkey-io/valkey-glide-php/commit/501c7dd24797cdea5289826592242c88ee39f27e

valkey_glide.stub.php
prateek-kumar-improving: Updated in this commit id: https://github.com/valkey-io/valkey-glide-php/commit/501c7dd24797cdea5289826592242c88ee39f27e

valkey_glide_cluster.stub.php
prateek-kumar-improving: Updated in this commit id: https://github.com/valkey-io/valkey-glide-php/commit/501c7dd24797cdea5289826592242c88ee39f27e

tests/ValkeyGlideTest.php
prateek-kumar-improving: The generic client() path returns true/false for OK responses regardless of OPT_REPLY_LITERAL — string conversion only applies to dedicated methods like clientPause(). Will add a proper reply-literal test when dedicated clientTracking()/clientCaching() methods land in #296 if required there.

tests/ValkeyGlideTest.php
prateek-kumar-improving: yes, routing is only present in cluster.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: No update required.

CHANGELOG.md
prateek-kumar-improving: No code update required.

valkey_glide_commands_3.c
prateek-kumar-improving: No required client side. No code change.

valkey_glide_core_common.c
prateek-kumar-improving: No code change required.

valkey_glide_commands_3.c
prateek-kumar-improving: No code change required.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: No code change required.

valkey_glide_commands.c
prateek-kumar-improving: No code change required.

tests/ValkeyGlideTest.php
prateek-kumar-improving: Updated in this commit id: https://github.com/valkey-io/valkey-glide-php/commit/1b9bd054298fdd6d1f34421a10d8f13676fb2205

tests/ValkeyGlideTest.php
prateek-kumar-improving: Will add when the dedicated `clientTracking()` method is implemented in #296.

tests/ValkeyGlideTest.php
prateek-kumar-improving: Acknowledged — will add negative tests for invalid option combinations when the typed `clientTracking()` wrapper is implemented. Noted in #296.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+* Add `CLIENT CACHING`, `CLIENT TRACKING`, and `CLIENT TRACKINGINFO` commands for standalone and cluster clients ([#293](https://github.com/valkey-io/valkey-glide-php/pull/293))
```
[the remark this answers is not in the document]
prateek-kumar-improving: Updated.

tests/ValkeyGlideClusterTest.php
```diff
@@ -148,6 +148,26 @@ public function testFailoverSucceeds()
         $this->markTestSkipped('FAILOVER not supported in cluster mode');
     }
 
+    /* CLIENT TRACKING/CACHING state tests don't work reliably in cluster mode
```
affonsov: client tracking and caching don't work reliably because the multiplexed connection that is why is not exposed through api

tests/ValkeyGlideClusterTest.php
```diff
@@ -148,6 +148,26 @@ public function testFailoverSucceeds()
...
+     * because tracking is connection-specific and commands may route to different
+     * nodes. The cluster-specific tests below verify the commands work. */
+    public function testClientTrackingInfoAfterEnable()
+    {
+        $this->markTestSkipped('CLIENT TRACKING state verification unreliable in cluster mode');
```
affonsov: you can clean up the tests, since we don't support client caching/tracking due the multiplexed connection for now

valkey_glide_commands_3.c
```diff
@@ -349,6 +349,12 @@ static enum RequestType determine_client_command_type(zval* args, int args_count
...
+            return ClientCaching;
+        } else if (strcasecmp(subcmd, "TRACKING") == 0) {
```
affonsov: missing clean up?

valkey_glide_core_common.c
```diff
@@ -317,7 +318,9 @@ int prepare_core_args(core_command_args_t* args,
 
         /* Message operations (no key, just arguments) */
         case BgSave:
+        case ClientCaching:
```
affonsov: missing clean up?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+* Add `CLIENT TRACKINGINFO` command for standalone and cluster clients. Add `CLIENT CACHING` and `CLIENT TRACKING` support via the generic `client()` method ([#293](https://github.com/valkey-io/valkey-glide-php/pull/293))
```
affonsov: what generic support means?

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,118 @@ public function testClientPauseBatch()
...
+
+    public function testClientTrackingInfoAfterEnable()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // Enable tracking via generic client() method
+        $this->valkey_glide->client('TRACKING', 'ON');
```
affonsov: we should not write tests like this, this is actually not testing the client side caching. you need to get the server_assisted PR first, because it will really enable to feature (client tracking) in the core.

We should avoid tests like this, because users can see this and think that they can use like that.

affonsov (changes requested):
We should first enable serverAssited so you can create proper tests
We also should avoid using custom commands on these cases, because for cluster tracking/caching will not work properly due the multiplexed connection

Please update the PR description

CHANGELOG.md
prateek-kumar-improving: It is for the generic PHPRedis compatible method called client() that supports multiple cases through the same method signature.

tests/ValkeyGlideTest.php
prateek-kumar-improving: Not removing them since we will need them later. Instead commenting them and adding a TO-DO with the github issue link.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated the comments in the test: https://github.com/valkey-io/valkey-glide-php/pull/293/changes/b0848ffe7f45292f6867b6c7f91b1e61da4acfa2

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Updated the message in the tests: https://github.com/valkey-io/valkey-glide-php/pull/293/changes/b0848ffe7f45292f6867b6c7f91b1e61da4acfa2

These tests in cluster will need to added because they are in the base class, PHP expects methods in the base class to be implemented here. There is a separate issue to find a way to remove all these skipped tests in cluster: https://github.com/valkey-io/valkey-glide-php/issues/287

valkey_glide_commands_3.c
prateek-kumar-improving: commented for now: https://github.com/valkey-io/valkey-glide-php/commit/b0848ffe7f45292f6867b6c7f91b1e61da4acfa2. Added a TO-DO to add back so that it is added back with caching task itself: https://github.com/valkey-io/valkey-glide-php/issues/287

CHANGELOG.md
prateek-kumar-improving: Updated the changelog to only client trackinginfo for now. Generic client method will be updated later.

prateek-kumar-improving:
> We should first enable serverAssited so you can create proper tests We also should avoid using custom commands on these cases, because for cluster tracking/caching will not work properly due the multiplexed connection
> 
> Please update the PR description

Updated the PR description, changelog, PR title to "Adding client trackinginfo command". Added TO-DOs everywhere so that these are added with this issue: #296. Also, mentioned all code that needs to be uncommented in the issue also.

valkey_glide_core_common.c
prateek-kumar-improving: Updated. Added a TO-DO with a comment to add later.

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +819,97 @@ public function testClientPauseBatch()
...
 
+    public function testClientTrackingInfoDefault()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // Default route in cluster mode returns per-node map: {node_address => tracking_info}
+        $result = $this->valkey_glide->clientTrackingInfo();
```
xShinnRyuu: Wrong default-route assumption. `CLIENT TRACKINGINFO` is `RouteBy::Random` in glide-core (`cluster_routing.rs:768`), not `AllPrimaries` like `MEMORY DOCTOR` / `LATENCY LATEST`. With no explicit route the command lands on a single random node, so the response is the flat `{flags, redirect, prefixes}` associative array, not a per-node map keyed by `node_address`. This is what is fataling CI:

```
Warning: Undefined array key "flags" in ValkeyGlideClusterTest.php on line 841
...
Fatal error: Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, null given in ValkeyGlideClusterTest.php:844
```

Suggested fix: mirror `testClientTrackingInfoWithRoute` here (single flat response), and drop the `foreach ($result as $nodeAddress => $info)` loop. If you specifically want to exercise the per-node shape by default, this is really a `testClientTrackingInfoDefault` = flat response test, and the per-node coverage already lives in `testClientTrackingInfoWithAllNodesRoute`. Also update the leading comment on this line to reflect the actual routing (`Default route in cluster mode returns a single flat map from a random node.`).

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +819,97 @@ public function testClientPauseBatch()
...
+
+    public function testClientTrackingInfoWithNullRoute()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // null route falls back to default (all nodes in cluster mode)
+        $result = $this->valkey_glide->clientTrackingInfo(null);
```
xShinnRyuu: Same routing bug as `testClientTrackingInfoDefault`. Passing `null` in the cluster handler simply skips setting `has_route` (see `valkey_glide_commands.c` `execute_client_tracking_info_command` around L1922 to L1927), which is behaviorally identical to omitting the argument. It does not mean `allNodes`. The response will be a flat single-node map, so `foreach ($result as $nodeAddress => $info)` will iterate `flags`, `redirect`, `prefixes` and every `$info[...]` lookup will fail.

Suggested fix: assert the flat shape here, matching `testClientTrackingInfoWithRoute`, and update the inline comment (`// null route falls back to default (all nodes in cluster mode)` is wrong: the default is `RouteBy::Random`, i.e. a single random node).

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +819,97 @@ public function testClientPauseBatch()
...
+
+    public function testClientTrackingInfoWithAllNodesRoute()
+    {
+        if (!$this->minVersionCheck('6.2.0')) {
+            $this->markTestSkipped('CLIENT TRACKINGINFO requires 6.2.0+');
+            return;
+        }
+
+        // AllNodes route returns per-node map, each node should be off by default
+        $result = $this->valkey_glide->clientTrackingInfo('allNodes');
```
xShinnRyuu, line 874: This one is correct: `allNodes` overrides the `RouteBy::Random` default, so the per-node map shape is what glide-core returns. Just leaving a note here so it does not get "fixed" together with the two above by accident: `testClientTrackingInfoWithAllNodesRoute` should stay as-is.

tests/ValkeyGlideClusterTest.php
```diff
@@ -804,6 +819,97 @@ public function testClientPauseBatch()
...
+        foreach ($result as $nodeAddress => $info) {
+            $this->assertIsString($nodeAddress);
+            $this->assertIsArray($info);
+            $this->assertArrayHasKey('flags', $info);
+            $this->assertArrayHasKey('redirect', $info);
+            $this->assertArrayHasKey('prefixes', $info);
+            $this->assertIsArray($info['flags']);
+            $this->assertTrue(is_int($info['redirect']) || is_long($info['redirect']));
+            $this->assertIsArray($info['prefixes']);
+            $this->assertTrue(in_array('off', $info['flags']) || array_key_exists('off', $info['flags']));
```
xShinnRyuu: Low-severity: `array_key_exists('off', $info['flags'])` is dead defensive coding. After glide-core's `ClientTrackingInfoReturnType` conversion (see `value_conversion.rs:1146` to `1176`), `flags` is always a `Value::Set`, which the extension decodes via the `Sets` case in `command_response.c:665` to `676`. That path uses `add_next_index_zval`, producing an integer-keyed indexed array. So `array_key_exists('off', ...)` is checking whether the integer key `"off"` exists (it never can), and `in_array('off', $info['flags'])` alone is the correct assertion. The same pattern is duplicated at lines 864 and 908, and at `tests/ValkeyGlideTest.php:167`. Consider dropping the `|| array_key_exists(...)` half everywhere for clarity.

valkey_glide.stub.php
```diff
@@ -2427,6 +2427,21 @@ public function clientPause(int $timeout, ?string $mode = null): ValkeyGlide|boo
...
+     *                                  - "flags" (array): A list of strings representing active tracking flags
+     *                                    (e.g., ["off"], ["on", "bcast"]).
+     *                                  - "redirect" (int): Client ID for redirection, or -1 if not set.
+     *                                  - "prefixes" (array): A list of strings representing key prefixes being
+     *                                    tracked. Empty array (not null) when no prefixes are tracked.
+     *                                  Returns false on failure.
+     *
+     * @see https://valkey.io/commands/client-trackinginfo
+     */
+    public function clientTrackingInfo(): ValkeyGlide|array|false;
```
xShinnRyuu, line 2443: Nit: the cluster stub accepts `mixed $route = null`, but the standalone stub takes no argument. This is a compatible narrowing per the PHP contract (LSP is respected: the cluster class overrides the parent method by adding an optional parameter, which is allowed), but it does mean code written against the cluster stub that unconditionally passes `null` will hit an `ArgumentCountError` when talking to a standalone. Two low-effort options:

1. Add `mixed $route = null` to the standalone signature and simply ignore it in the C handler (matches how the parameter is already ignored on standalone semantically).
2. Leave the signature and add a short docblock note here that this method takes no route on standalone.

Either is fine; the current state just leaves a minor pit for cross-client callers.

valkey_glide_commands_3.c
```diff
@@ -349,6 +349,10 @@ static enum RequestType determine_client_command_type(zval* args, int args_count
             return ClientInfo;
         } else if (strcasecmp(subcmd, "NO-EVICT") == 0) {
             return ClientNoEvict;
+            // TODO(#296): Uncomment CLIENT CACHING and CLIENT TRACKING
```
xShinnRyuu: Cosmetic: the `TODO(#296): Uncomment CLIENT CACHING and CLIENT TRACKING` comment is nested inside the `NO-EVICT` `else if` branch and reads as attached to `NO-EVICT` rather than being a top-level dispatcher note. Move it above the `TRACKINGINFO` branch (or below the closing `}` of the outer `if`) so it clearly annotates the missing subcommands, not `NO-EVICT`.

tests/ValkeyGlideTest.php
```diff
@@ -3113,6 +3113,120 @@ public function testClientPauseBatch()
...
+
+        // Default state: tracking off, no redirect
+        $this->assertTrue(in_array('off', $info['flags']) || array_key_exists('off', $info['flags']));
+        $this->assertEquals(-1, $info['redirect']);
+        $this->assertEmpty($info['prefixes']);
+    }
+
+    // TODO: Re-enable CLIENT TRACKING/CACHING tests once server-assisted client-side caching is implemented.
+    //       See: https://github.com/valkey-io/valkey-glide-php/issues/296
+    // public function testClientTrackingInfoAfterEnable()
```
xShinnRyuu: If we are leaving TODOs lets be a little more consistent with how we add them in as comments, so that we dont end up missing them or forgetting about them once the follow up task is done.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: This test was failing. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/edd882e

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: This test was failing. Updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/edd882e

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: No code change required.

valkey_glide.stub.php
prateek-kumar-improving: Standalone does not have route or any mention of route. This is the pattern for all implemented commands.  
If this is actually adding value, it has to updated for all commands in a separate issue. Please create a separate issue for it and it will need to analyzed separately.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Removed the || array_key_exists('off', $info['flags']) from all active assertions. Since flags is always an indexed array, in_array() alone is correct. Updated in this commit: https://github.com/valkey-io/valkey-
glide-php/commit/b36a144

valkey_glide_commands_3.c
prateek-kumar-improving: It was moved there because lint fails if not placed before the closing `}`

tests/ValkeyGlideTest.php
prateek-kumar-improving: There is a issue link and it is all mentioned in the issue as well. Currently this is how it is being done in the PHP repo. So, it is consistent for PHP with TODO then some description and an issue link. Is there any specific way you are recommending.

The changes to be made are also mentioned in the issue.

xShinnRyuu: There are 2 different styles of TODO in this one PR. Below is what i mean by consistency

```
// TODO(#296): Uncomment CLIENT CACHING and CLIENT TRACKING
```

```
// TODO: Re-enable CLIENT TRACKING/CACHING tests once server-assisted client-side caching is implemented.
    //       See: https://github.com/valkey-io/valkey-glide-php/issues/296
```

prateek-kumar-improving: Updated all TODOs to use the consistent `// TODO(#296)`: format.
Updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/75d8d8f

tests/ValkeyGlideClusterTest.php
affonsov: Client Side caching tests are located https://github.com/valkey-io/valkey-glide-php/blob/main/tests/ClientSideCacheTest.php
Any related test or TODO should be moved to there.

valkey_glide_commands_3.c
affonsov: We shouldn't expose this or neither add those in the TODO for now, we are not exposing this command as is in the glide for now

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: I understand, but client command will be separate for PHPRedis compatibility. And clientTrackingInfo is a separate command in itself.

valkey_glide_commands_3.c
prateek-kumar-improving: We will need to add those for client to work properly because of PHPRedis compatibility.

prateek-kumar-improving: As discussed we are not implementing these APIs. Removed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/293/changes/d0468915f59efdd2c71a70e1b23bbfd8d0c0e073

valkey_glide_core_common.c
prateek-kumar-improving: Removed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/293/changes/d0468915f59efdd2c71a70e1b23bbfd8d0c0e073

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: As discussed TO-DO's removed. Removed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/d0468915f59efdd2c71a70e1b23bbfd8d0c0e073

xShinnRyuu (approved):

affonsov (approved):

PR #309, opened 2026-08-13
feat(php): Implement monitor command

prateek-kumar-improving:
### Summary

Implement the MONITOR command for the PHP standalone client as a dedicated `ValkeyGlideMonitor` class, mirroring the design used by the other Valkey GLIDE clients (Go, Python, Node, Java). It uses the existing `create_monitor_client` / `close_monitor_client` FFI functions from glide-core's Rust layer, opening a dedicated connection for monitor streaming.

The client delivers **structured** `ValkeyGlideMonitorLine` records (not a raw text string) and supports **both** consumption styles — a pull API and a callback loop — matching the other clients.

### Issue link

### Design

Following reviewer feedback, MONITOR is implemented as a **separate client class** rather than a method on `ValkeyGlide`. This matches Go (`NewMonitorClient`), Python (`MonitorClient.create`), Node (`GlideMonitorClient.create`), and Java (`MonitorClient`), and keeps the main `ValkeyGlide` object free of monitor-specific state.

```php
$monitor = new ValkeyGlideMonitor(addresses: [['host' => 'localhost', 'port' => 6379]]);

// Pull API (parity with GetMonitorMessage / get_monitor_message)
$line = $monitor->getMonitorMessage(timeout: 5.0);   // blocking; null on timeout
$line = $monitor->tryGetMonitorMessage();            // non-blocking; null if empty

// Structured line
echo $line->command;      // "SET"
print_r($line->args);     // ["key", "value"]
echo (string)$line;       // "1339877440.333333 [0 127.0.0.1:6379] \"SET\" \"key\" \"value\""

// Callback loop (PHP-specific convenience; non-null return exits)
$monitor->listen(function (ValkeyGlideMonitorLine $line) {
    if (strcasecmp($line->command, 'SET') === 0) return true;
});

echo $monitor->getDroppedCount();  // lines dropped if the queue overflowed
$monitor->close();
```

- New `ValkeyGlideMonitor` class (standalone only, not cluster).
- New `ValkeyGlideMonitorLine` value object with decoded fields: `timestamp`, `db`, `clientAddr`, `command`, `args[]`, plus `__toString()` for the PHPRedis-compatible text form.
- Pull API: `getMonitorMessage(?float $timeout)` (blocking) and `tryGetMonitorMessage()` (non-blocking).
- Callback API: `listen(callable $cb)` — blocking loop, non-null return exits (PHPRedis `subscribe()`-style).
- `getDroppedCount()` surfaces queue-overflow losses (preserved across `close()`).
- `close()` stops the dedicated connection and releases queue state.
- `ValkeyGlide` no longer carries any monitor state (`monitor()`, `connection_request_bytes`, and `in_monitor_mode` were removed).

### Implementation

**Architecture**:
1. `ValkeyGlideMonitor::__construct` parses standalone connection params, serializes them to a protobuf connection request, and calls `create_monitor_client()` (opens a dedicated TCP connection + sends MONITOR).
2. The Rust core invokes the C callback (`valkey_glide_monitor_callback`) from a background thread for each parsed line. The callback stores the **raw** fields (timestamp, db, client addr, command, decoded args) in a thread-safe queue (mutex + condvar) — no Zend APIs are touched on the background thread.
3. PHP-thread consumers build a `ValkeyGlideMonitorLine` object from a dequeued node: the pull methods dequeue on demand; `listen()` drains in a loop and invokes the callback.
4. On `close()` / object destruction, `close_monitor_client()` stops the background task first, then callback state is released.

**Key files**:
- `valkey_glide_monitor.c` / `.h` — the `ValkeyGlideMonitor` + `ValkeyGlideMonitorLine` classes, constructor, pull/callback methods, object handlers.
- `valkey_glide_monitor_common.c` / `.h` — shared core: background callback, JSON-arg decoding, thread-safe queue, native registry, `dequeue`.
- `valkey_glide_monitor.stub.php` — PHP API stubs for both classes.
- `valkey_glide.c` / `common.h` — removal of monitor state from the main client; MINIT registration of the new classes.

### Limitations

- MONITOR is standalone only (not available on `ValkeyGlideCluster`), consistent with GLIDE across all languages.
- MONITOR degrades server performance; intended for debugging only.
- Commands issued concurrently with entering monitor mode may not be captured — a shared limitation across all GLIDE clients (the Rust FFI drops lines until the stream is active).

### Testing

Integration tests in `tests/ValkeyGlideMonitorTest.php`:

| Test | Verifies |
|------|----------|
| `testMonitorRequiresCallable` | `listen()` rejects a non-callable |
| `testMonitorPullReturnsStructuredLine` | Pull API returns a structured `ValkeyGlideMonitorLine` (fields + `__toString`) |
| `testMonitorTryGetIsNonBlocking` | `tryGetMonitorMessage()` does not block |
| `testMonitorCapturesSetCommand` | SET from another client captured |
| `testMonitorCapturesGetCommand` | GET from another client captured |
| `testMonitorOutputFormat` | Timestamp + db + addr format |
| `testMonitorCallbackReturnExits` | Non-null return exits `listen()` |
| `testMonitorCapturesMultipleCommands` | SET/GET/DEL all captured |
| `testMonitorCloseAfterMonitorIsSafe` | Clean shutdown after monitor |
| `testMonitorNotAvailableOnCluster` | No `monitor()` on `ValkeyGlide`/`ValkeyGlideCluster`; monitor classes exist |
| `testMonitorOutputContainsArgs` | Key + value in output |

The subprocess-based capture tests use the same pattern as the existing pub/sub tests, with sync files for coordination.

**Build verified**: compiles cleanly with `-Werror`; all monitor tests pass locally (macOS, PHP 8.5, Apple Silicon).

### Checklist


valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,551 @@
...
+    msg->next     = NULL;
+
+    // Add to queue (thread-safe) with bounded depth
+    mutex_lock(&info->queue_mutex);
+    if (info->queue_depth >= MONITOR_QUEUE_MAX_DEPTH) {
+        // Queue full — drop the message to prevent unbounded memory growth
+        mutex_unlock(&info->queue_mutex);
+        free(line_buf);
+        free(msg);
+        return;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 81bfceb08bab67adcf9bc86950cc5a262bba1d6e

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,551 @@
...
+                                   double         timestamp,
+                                   int64_t        db,
+                                   const uint8_t* client_addr,
+                                   int64_t        client_addr_len,
+                                   const uint8_t* command_ptr,
+                                   int64_t        command_len,
+                                   const uint8_t* args_json,
+                                   int64_t        args_json_len) {
+    // Look up callback info using native registry (no Zend APIs)
+    monitor_callback_info* info = native_registry_find(client_ptr);
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 5b24347ace2c5009d4db7c07e0d55d7f81eba133

valkey_glide_monitor_common.h
```diff
@@ -0,0 +1,71 @@
...
+    monitor_message* queue_tail;
+    size_t           queue_depth;
+    mutex_t          queue_mutex;
+    cond_t           queue_cond;
+    bool             in_monitor_mode;
+} monitor_callback_info;
+
+// Maximum number of messages allowed in the queue before dropping new ones.
+// Prevents unbounded memory growth if the PHP callback is slower than traffic.
+#define MONITOR_QUEUE_MAX_DEPTH 10000
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 81bfceb08bab67adcf9bc86950cc5a262bba1d6e

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,573 @@
...
+    msg->next     = NULL;
+
+    // Add to queue (thread-safe) with bounded depth
+    mutex_lock(&info->queue_mutex);
+    if (info->queue_depth >= MONITOR_QUEUE_MAX_DEPTH) {
+        // Queue full — drop the message to prevent unbounded memory growth
+        mutex_unlock(&info->queue_mutex);
+        free(line_buf);
+        free(msg);
+        return;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 81bfceb08bab67adcf9bc86950cc5a262bba1d6e

valkey_glide.c
```diff
@@ -1089,6 +1099,21 @@ static int valkey_glide_create_connection(valkey_glide_object* valkey_glide,
...
 
+    /* Store connection request bytes for monitor (needs its own dedicated connection).
+     * We regenerate the bytes here before the config is cleaned up. */
+    {
+        size_t   mon_len   = 0;
+        uint8_t* mon_bytes = create_connection_request(
+            &mon_len, &client_config, VALKEY_GLIDE_PERIODIC_CHECKS_DISABLED, false, false);
+        if (mon_bytes && mon_len > 0) {
+            valkey_glide->connection_request_bytes = mon_bytes;
+            valkey_glide->connection_request_len   = mon_len;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 81bfceb08bab67adcf9bc86950cc5a262bba1d6e

valkey_glide.c
```diff
@@ -1089,6 +1099,21 @@ static int valkey_glide_create_connection(valkey_glide_object* valkey_glide,
...
+        } else {
+            valkey_glide->connection_request_bytes = NULL;
+            valkey_glide->connection_request_len   = 0;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 81bfceb08bab67adcf9bc86950cc5a262bba1d6e

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,601 @@
...
+    }
+    if (info->queue_tail) {
+        info->queue_tail->next = msg;
+    } else {
+        info->queue_head = msg;
+    }
+    info->queue_tail = msg;
+    info->queue_depth++;
+    cond_signal(&info->queue_cond);
+    mutex_unlock(&info->queue_mutex);
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 81bfceb08bab67adcf9bc86950cc5a262bba1d6e

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,666 @@
...
+                if (client_ptrs) {
+                    client_ptrs[i] = entry->client_ptr;
+                }
+                entry = entry->next;
+            }
+            if (!client_ptrs) {
+                // Allocation failed — skip the proactive close loop below.
+                // Each monitor's blocking loop will still notice is_active is
+                // false (via its timed wait) and exit on its own.
+                client_count = 0;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 67407e34c5aa8577db716182f5f2792f70183342

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,686 @@
...
+// Native registry: add entry (thread-safe)
+static void native_registry_add(uintptr_t client_ptr, monitor_callback_info* info) {
+    monitor_registry_entry* entry =
+        (monitor_registry_entry*) malloc(sizeof(monitor_registry_entry));
+    if (!entry) {
+        // Allocation failure — the callback info remains registered in the
+        // Zend-side hashtable, but the background thread won't be able to
+        // find it via the native registry, so monitor lines will simply not
+        // be delivered rather than crashing.
+        return;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 67407e34c5aa8577db716182f5f2792f70183342

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,686 @@
...
+                if (client_ptrs) {
+                    client_ptrs[i] = entry->client_ptr;
+                }
+                entry = entry->next;
+            }
+            if (!client_ptrs) {
+                // Allocation failed — skip the proactive close loop below.
+                // Each monitor's blocking loop will still notice is_active is
+                // false (via its timed wait) and exit on its own.
+                client_count = 0;
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 67407e34c5aa8577db716182f5f2792f70183342

tests/scripts/monitor_multi_capture.php
```diff
@@ -0,0 +1,73 @@
...
+// Monitor multi-capture script for testing
+// Usage: php monitor_multi_capture.php <host> <port> <sync_file> <result_file> <prefix> <expected_count>
+//
+// Captures all monitor lines containing the given prefix, up to expected_count matches,
+// then writes them all (newline-separated) to result_file.
+//
+// NOTE: The "ready" signal is written just before monitor() because monitor() is
+// blocking and create_monitor_client() completes synchronously within it. The parent
+// test uses a small delay after receiving the signal to ensure the monitor loop has
+// started receiving events.
```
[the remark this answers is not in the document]
prateek-kumar-improving: code fixed in this commit 454107351ab19b2ad8ee99f6e905c16d906474f2

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,640 @@
...
+
+    // Register the PHP callback using the monitor client pointer as key
+    // (not the main client pointer, since the monitor has its own connection).
+    // If registration fails (e.g. allocation failure), the background
+    // producer would never be able to find this monitor and every event
+    // would be silently discarded — close the connection we just opened and
+    // fail loudly instead of entering a blocking loop that can never
+    // receive anything.
+    monitor_callback_info* info =
+        php_register_monitor_callback((uintptr_t) monitor_client_ptr, callback, ZEND_THIS);
```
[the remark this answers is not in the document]
prateek-kumar-improving: This is an existing rust core issue: https://github.com/valkey-io/valkey-glide/issues/6781. It should be fixed for all languages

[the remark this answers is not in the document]
prateek-kumar-improving: This is an existing rust core issue: https://github.com/valkey-io/valkey-glide/issues/6781. It should be fixed for all languages

valkey_glide.stub.php
```diff
@@ -2222,6 +2222,39 @@ public function ltrim(string $key, int $start, int $end): ValkeyGlide|bool;
...
+     * $valkey_glide->monitor(function ($valkey_glide, $command) {
+     *     echo "$command\n";
+     *
+     *     // Stop monitoring after seeing a SET command
+     *     if (str_contains($command, '"SET"') || str_contains($command, '"set"')) {
+     *         return true;
+     *     }
+     * });
+     */
+    public function monitor(callable $cb): bool;
```
affonsov: Every other client implements MONITOR as a *dedicated object* created from a config — Go `NewMonitorClient(cfg, callback)`, Python `MonitorClient.create(config, callback)`, Node `GlideMonitorClient.create(config, callback?)`, Java `MonitorClient` — each delivering a **structured** `MonitorLine`/`MonitorMsg` (`{timestamp, db, clientAddr, command, args[]}`) with args **decoded** into a list, and each offering **both** a callback and a pull API (`getNextMessage`/`GetMonitorMessage`/`get_monitor_message`) with the callback optional.

This PR instead hangs `monitor()` off the main client, blocks with a *required* callback (exit-on-non-null-return), and hands the callback a **reconstructed text string** with args left as verbatim JSON tokens — no structured fields, no pull API. The blocking-callback model is reasonable for PHP (it's synchronous and matches this extension's own pub/sub), but the string-instead-of-`MonitorLine` surface means PHP users must regex-parse to recover the command/args that every other client returns structured. 

Just want to understand why we need to diverge the design applied to the other clients

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,640 @@
...
+            }
+            break;
+        }
+
+        if (dropped > 0) {
+            char overflow_line[128];
+            int  overflow_len =
+                snprintf(overflow_line,
+                         sizeof(overflow_line),
+                         "*** MONITOR OVERFLOW: %zu command(s) dropped (queue full) ***",
```
affonsov: Surfacing dropped-command counts instead of silently discarding is genuinely better than Go's silent `select { default }` drop — good call. But this synthetic line `*** MONITOR OVERFLOW: N command(s) dropped (queue full) ***` is delivered through the same callback as real lines and doesn't match the documented format (`timestamp [db addr] "COMMAND" ...`) — which your own docblock example and the tests parse with `/^\d+\.\d+/`. A user parsing lines by that shape will choke on it. Either document that a non-conforming overflow-notice line can appear, or signal overflow out-of-band / with a documented prefix, so line parsers aren't broken.

valkey_glide.c
```diff
@@ -1089,6 +1099,28 @@ static int valkey_glide_create_connection(valkey_glide_object* valkey_glide,
 
     free_connection_response((ConnectionResponse*) conn_resp);
 
+    /* Store connection request bytes for monitor (needs its own dedicated connection).
```
affonsov: Connection-request bytes are regenerated and retained on *every* `connect()`, for every client, whether or not monitor is ever used.

This runs a second `create_connection_request()` serialization on the connect path and keeps a copy of the connection request (which includes credentials) resident on the object for the client's whole lifetime, purely to support a debug command most clients never call. The zero-on-free is good, but it widens the window credentials sit in heap for all users and adds per-connect work. Consider retaining/regenerating lazily (only when `monitor()` is actually invoked) or gating it behind a flag. Not a blocker, but it's a cost every connection pays for a rarely-used feature.

tests/ValkeyGlideMonitorTest.php
```diff
@@ -0,0 +1,717 @@
...
+    /**
+     * Wait for a listener to acknowledge that its dedicated MONITOR connection
+     * is established and receiving events.
+     *
+     * The listener writes $syncFile from its first monitor callback, not before
+     * entering monitor(). Sending PING probes until that file appears therefore
+     * creates an explicit activation handshake: the acknowledgement cannot
+     * exist until Valkey has delivered at least one record through MONITOR.
+     */
+    protected function waitForMonitorActivation(string $syncFile, int $timeoutSeconds = 5)
```
affonsov, line 813: Document the startup-drop limitation, and note that these tests can't catch it.

1. This is a **shared, documented limitation**, not a PHP regression — Go/Python/Node register after create too, and the Rust FFI itself drops lines while its `ptr_cell == 0` and its doc says *"callers should not rely on receiving lines issued concurrently with create returning."*
2. The suite **structurally can't exercise it**: `waitForMonitorActivation()` PINGs until the first callback fires and only *then* sends the real command, so the drop window is never tested.

Please add a one-line caveat to the `monitor()` docblock that commands issued concurrently with entering monitor mode may not be captured, matching the other clients' behavior.

valkey_glide.stub.php
currantw: +1

currantw (commented):
Did a high-level review. I agree with @affonsov that we should have a separate Monitor client in order to match other Valkey GLIDE clients.

valkey_glide.stub.php
prateek-kumar-improving: It is for PHPRedis compatibility. I considered other client designs, but decided to go with the best design for PHPRedis compatibility.

valkey_glide_monitor_common.c
prateek-kumar-improving: Code fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/95a256d

valkey_glide.c
prateek-kumar-improving: Code fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/95a256d

tests/ValkeyGlideMonitorTest.php
prateek-kumar-improving: Code fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/95a256d

affonsov (approved):
Conditionally approving
Looks like CI need some fixes and the PR summary is outdated, please update that

prateek-kumar-improving:
> Conditionally approving Looks like CI need some fixes and the PR summary is outdated, please update that

Updated PR summary.

.gitignore
```diff
@@ -97,6 +97,8 @@ valkey_glide_arginfo.h
 valkey_glide_cluster_arginfo.h
 valkey_glide_cluster_legacy_arginfo.h
 valkey_glide_legacy_arginfo.h
+valkey_glide_monitor_arginfo.h
+valkey_glide_monitor_legacy_arginfo.h
```
currantw, line 101: Why are these necessary? Are these created by PHP during build?

valkey_glide.c
```diff
@@ -16,6 +16,8 @@
 #include "valkey_glide_cluster_arginfo.h"  // Include generated arginfo header
 #include "valkey_glide_commands_common.h"
 #include "valkey_glide_core_common.h"
+#include "valkey_glide_monitor.h"  // Include ValkeyGlideMonitor class
```
currantw: Nit. Comment doesn't add anything 🤷 


```suggestion
#include "valkey_glide_monitor.h"
```

valkey_glide_monitor.stub.php
```diff
@@ -0,0 +1,152 @@
...
+ * A single line received from the MONITOR command, in structured form.
+ *
+ * Mirrors the MonitorLine / MonitorMsg types exposed by the other Valkey GLIDE
+ * clients (Go, Python, Node, Java) so that PHP users receive the same decoded
+ * fields instead of a raw text line they must parse themselves.
+ */
+final class ValkeyGlideMonitorLine
+{
+    /** Server timestamp (seconds with microsecond fraction). */
+    public float $timestamp;
```
currantw: What does it mean to have a "microsecond fraction"? Is this not just normal floating-point seconds?

valkey_glide_monitor.stub.php
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Render the line in the PHPRedis-compatible text format:
+     * "timestamp [db addr] \"COMMAND\" \"arg1\" \"arg2\"".
+     *
+     * @return string The formatted monitor line.
+     */
+    public function __toString(): string
+    {
+    }
```
currantw: Since PHPRedis does not support the monitor command, what does it mean to be "PHPRedis-compatible text format" in this context?

valkey_glide_monitor.stub.php
```diff
@@ -0,0 +1,152 @@
...
+ * own connection (separate from any ValkeyGlide client) and delivers structured
+ * ValkeyGlideMonitorLine records. It is standalone-only; cluster mode is not
+ * supported, consistent with the other Valkey GLIDE clients.
+ *
+ * Warning: MONITOR is a debugging tool that degrades server performance. Do not
+ * use it in production for extended periods.
+ *
+ * Two consumption styles are supported:
+ *  - Pull: call getMonitorMessage()/tryGetMonitorMessage() from your own loop.
+ *  - Callback: call listen() with a callback (PHP-specific convenience).
```
currantw: We should link the documentation here:
https://glide.valkey.io/how-to/monitoring/monitor-command/

We should also align terminology with the documentation, which calls these two modes "Callback" and "Queue" (rather than "Pull").

valkey_glide_monitor.stub.php
```diff
@@ -0,0 +1,152 @@
...
+     * This is a PHP-specific convenience over getMonitorMessage(). Return any
+     * non-null value from the callback to exit the loop.
+     *
+     * @param callable $cb Receives one ValkeyGlideMonitorLine argument per line.
+     *
+     * @return bool True when the loop exits normally.
+     *
+     * @throws ValkeyGlideException If the callback is not callable.
+     */
+    public function listen(callable $cb): bool;
```
currantw, line 117: In the other clients, the monitor client can be _initialized_ with a callback method, which is called for the lifetime of the monitor client. Why is the PHP client implemented differently, by instead providing only for _runtime_ specification of a callback?

valkey_glide_monitor.stub.php
```diff
@@ -0,0 +1,152 @@
...
+     * @param array|null      $credentials      ['username' => string, 'password' => string, ...].
+     * @param int             $read_from        Read-from strategy (ValkeyGlide::READ_FROM_*).
+     * @param int|null        $request_timeout  Request timeout in milliseconds.
+     * @param array|null      $reconnect_strategy Reconnect backoff configuration.
+     * @param int|null        $database_id      Database index to select.
+     * @param string|null     $client_name      Client name reported to the server.
+     * @param string|null     $client_az        Availability zone hint.
+     * @param array|null      $advanced_config  Advanced configuration (e.g. TLS config).
+     * @param bool|null       $lazy_connect     Defer the connection until first use.
+     * @param mixed           $context          Stream context for TLS.
```
currantw: Some of the options (listed below), do not appear to do anything for monitor client? If so, I think they can be removed.

- `read_from`
- `request_timeout` 
- `reconnect_strategy`
- `client_az`
- `advanced_config`
- `lazy_connect`

I am not so sure about `client_name`. I guess it could technically be applicable, but without a clear use case, probably better to also omit? 🤷

valkey_glide_monitor.stub.php
```diff
@@ -0,0 +1,152 @@
...
+
+    /**
+     * Number of monitor lines dropped because the internal queue was full.
+     *
+     * Consult this to detect whether the stream was complete. The count is
+     * cumulative for the life of the client.
+     *
+     * @return int The number of dropped lines.
+     */
+    public function getDroppedCount(): int;
```
currantw, line 127: Do other languages expose a similar method/handle dropped lines the same?

valkey_glide_monitor_common.h
```diff
@@ -0,0 +1,97 @@
...
+// Stores the RAW decoded fields delivered by the background Rust thread rather
+// than a pre-formatted string. The PHP-side consumer builds either a
+// ValkeyGlideMonitorLine object (pull/callback) or a formatted string
+// (__toString) from these fields. Keeping raw fields avoids fragile C-side
+// string reconstruction and JSON re-escaping.
+typedef struct monitor_message {
+    double                  timestamp;    // Server timestamp (seconds.microseconds)
+    int64_t                 db;           // Database index
+    char*                   client_addr;  // "host:port" (owned, may be NULL)
+    char*                   command;      // Command name (owned, may be NULL)
```
currantw: Under what conditions would the command be `null` here? I know that the C# client assumed it is always populated.

valkey_glide_monitor_common.h
```diff
@@ -0,0 +1,97 @@
...
+// Returns a node the caller owns (must free with valkey_glide_monitor_free_message),
+// or NULL on timeout / inactive / empty.
+monitor_message* valkey_glide_monitor_dequeue(monitor_callback_info* info, long timeout_ms);
+
+// Maximum number of messages allowed in the queue before dropping new ones.
+// Prevents unbounded memory growth if the consumer is slower than traffic.
+// Drops are not silent: valkey_glide_monitor_callback() records them in
+// monitor_callback_info.dropped_count, which is surfaced to the user via
+// ValkeyGlideMonitor::getDroppedCount().
+#define MONITOR_QUEUE_MAX_DEPTH 10000
```
currantw, line 61: Is this how other clients handle drops/queue overflow?

valkey_glide_pubsub_common.h
```diff
@@ -75,6 +75,10 @@ void mutex_destroy(mutex_t* m);
 // Condition variable wrapper functions
 void cond_init(cond_t* c);
 void cond_wait(cond_t* c, mutex_t* m);
+// Returns 0 on success (woken by signal), a positive errno-style value on
+// timeout or failure (e.g. ETIMEDOUT), so callers can distinguish a timed-out
+// wait from a real wakeup if they need to.
+int  cond_timedwait(cond_t* c, mutex_t* m, unsigned int timeout_ms);
```
currantw, line 85: Why is this being added to the pub/sub commands? If these are shared, should we extract them to somewhere shared? Happy to defer to future issue if you want to raise it and tag with a TODO here?

valkey_glide_pubsub_common.c
```diff
@@ -62,6 +63,28 @@ void cond_wait(cond_t* c, mutex_t* m) {
...
+#ifdef _WIN32
+    BOOL ok = SleepConditionVariableCS(c, m, timeout_ms);
+    return ok ? 0 : 1;
+#else
+    struct timespec ts;
+    if (clock_gettime(CLOCK_REALTIME, &ts) != 0) {
+        // Failed to read the clock — fall back to an unbounded wait rather
+        // than using an uninitialized/garbage deadline.
+        pthread_cond_wait(c, m);
+        return 0;
```
currantw: Silently falling back to an unbounded wait – is that a good idea? 🤔

valkey_glide_pubsub_common.c
```diff
@@ -62,6 +63,28 @@ void cond_wait(cond_t* c, mutex_t* m) {
...
+    }
+    ts.tv_sec += timeout_ms / 1000;
+    ts.tv_nsec += (timeout_ms % 1000) * 1000000L;
+    if (ts.tv_nsec >= 1000000000L) {
+        ts.tv_sec++;
+        ts.tv_nsec -= 1000000000L;
+    }
```
currantw: When would this happen in practice? Can you give a numerical example perhaps?

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,571 @@
...
+        efree(info);
+    }
+}
+
+// Decode a single JSON string token into a newly malloc'd, null-terminated
+// buffer of raw bytes. `*pp` must point at the opening quote; on success it is
+// advanced past the closing quote. Returns NULL on allocation failure or if the
+// token is not a well-formed string. Runs on the background thread — native
+// only, no Zend APIs.
+static char* decode_json_string(const char** pp, const char* end) {
```
currantw: This is quite a complex piece of logic!

- What have you done to validate it under all reasonable conditions?
- Is there not a library or existing helper that could aid with this?
- Would it be possible to break this up using smaller functions at all?
- Would comments help to make this any more understandable?

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,571 @@
...
+
+    out[olen] = '\0';
+    *pp       = p;
+    return out;
+}
+
+// Parse a JSON array of strings ["a","b",...] into a newly-allocated char**
+// array of decoded, null-terminated C strings. Sets *out_count. Returns NULL
+// (count 0) if there are no elements. Native only — no Zend APIs.
+static char** parse_json_string_array(const uint8_t* args_json,
```
currantw: Similar to previous comment, I think we need unit tests to validate this logic, as it is quite complex!

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,571 @@
...
+    while (entry) {
+        if (entry->client_ptr == client_ptr) {
+            info = entry->info;
+            break;
+        }
+        entry = entry->next;
+    }
+
+    if (!info) {
+        goto unlock_registry;
```
currantw: You don't often see `goto` statements "in the wild!" Some developers really think these make code harder to follow and reason about. Did you consider other structures? Jumping to statements _after_ the usual `return` statement in particular seems strange to me – not sure I've ever seen that before.

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,571 @@
...
+
+        if (command_ptr && command_len > 0) {
+            msg->command = (char*) malloc((size_t) command_len + 1);
+            if (msg->command) {
+                memcpy(msg->command, command_ptr, (size_t) command_len);
+                msg->command[command_len] = '\0';
+            }
+        }
+
+        msg->args = parse_json_string_array(args_json, args_json_len, &msg->args_count);
```
currantw: Hmmm. In the C# client, we seem to receive the arguments as an array of strings. Why do we seem to be parsing JSON here? 🤔

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,571 @@
...
+        }
+    } else {
+        // Bounded wait. Loop in <=1s slices so PHP interruptions are honored
+        // by the caller between calls; here we just wait up to timeout_ms.
+        long remaining = timeout_ms;
+        while (!info->queue_head && info->is_active && remaining > 0) {
+            long slice = remaining < 1000 ? remaining : 1000;
+            (void) cond_timedwait(&info->queue_cond, &info->queue_mutex, slice);
+            remaining -= slice;
+        }
```
currantw: Is there a better way to wait a specified amount of time? What is the reason we need to break this up into "slices"?

tests/ValkeyGlideMonitorTest.php
```diff
@@ -0,0 +1,810 @@
...
+        $this->assertTrue($found !== null, 'Pull API should capture the SET command');
+        if ($found !== null) {
+            $this->assertTrue(is_float($found->timestamp), 'timestamp should be a float');
+            $this->assertTrue(is_int($found->db), 'db should be an int');
+            $this->assertTrue(is_string($found->clientAddr), 'clientAddr should be a string');
+            $this->assertTrue(in_array($key, $found->args, true), 'args should contain the key');
+            $this->assertTrue(in_array($val, $found->args, true), 'args should contain the value');
+            // __toString renders the PHPRedis-compatible line format.
+            $this->assertRegex('/^\d+\.\d+\s+\[\d+\s+[\d\.:]+\]/', (string)$found, 'toString format');
+        }
```
currantw, line 144: Can't we be more specific?
- Can we check timestamp from Valkey before running and verify that command timestamp is greater?
- Can we verify database is zero?
- Can we construct the client address string and verify exactly?
- Don't we know the exact command and arguments?
- Don't we know the exact string format?

tests/ValkeyGlideMonitorTest.php
```diff
@@ -0,0 +1,810 @@
...
+        $val = 'PULL_VAL_' . uniqid();
+
+        // Give the dedicated monitor connection a moment to activate.
+        $action = new ValkeyGlide();
+        $action->connect(addresses: [['host' => $this->getHost(), 'port' => $this->getPort()]]);
+        $action->set($key, $val);
+        $action->close();
+
+        $found = null;
+        for ($i = 0; $i < 20; $i++) {
```
currantw: - Maybe add a comment here on why this is needed? (I assume because other clients may also call methods on this server?)
- Why did you pick 20 as the max number to check?
- Could we wait for some specified period instead of a specified number of messages? Might that be less flaky?

tests/ValkeyGlideMonitorTest.php
```diff
@@ -0,0 +1,810 @@
...
+        );
+
+        // Drain any startup traffic quickly, then confirm a null is returned
+        // promptly once the queue is empty (non-blocking behavior).
+        $start = microtime(true);
+        for ($i = 0; $i < 100; $i++) {
+            if ($monitor->tryGetMonitorMessage() === null) {
+                break;
+            }
+        }
```
currantw: While not just continue the iteration until condition is reached with `while`? If worried about timeout, why not make that the condition? Seems more robust than a specific number of messages 🤷

tests/scripts/monitor_listener.php
currantw, line 1: Why is this needed? How is this run?

tests/scripts/monitor_multi_capture.php
currantw, line 1: Why is this needed? How is this run?

tests/ValkeyGlideMonitorTest.php
```diff
@@ -0,0 +1,810 @@
...
+            'tryGetMonitorMessage should not block waiting for messages'
+        );
+        $monitor->close();
+    }
+
+    /**
+     * Test that monitor captures SET commands from another client.
+     * Uses subprocess approach (same pattern as pubsub tests).
+     */
+    public function testMonitorCapturesSetCommand()
```
currantw, line 222: Why isn't the approach above sufficient – why do we need to *also* use this subprocess approach for testing?

tests/ValkeyGlideMonitorTest.php
```diff
@@ -0,0 +1,810 @@
...
+        $action_client->del($unique_prefix . '_key');
+        $action_client->close();
+
+        $timeout = time() + 5;
+        while (!file_exists($result_file) && time() < $timeout) {
+            usleep(100000);
+        }
+
+
+        // Terminate subprocess if it hasn't exited
```
currantw, line 585: Nit. Extra line.

currantw (commented):
Reviewed to ff3de2e.

valkey_glide_monitor.stub.php
prateek-kumar-improving: Other clients can run your callback on a background thread for the client's lifetime; PHP cannot safely call user code off the main thread, so instead of pretending otherwise via a constructor callback, PHP exposes the callback on `listen()`, which is the main-thread loop that actually drives it.

If cross-client familiarity matters in this case, we could also accept an optional `$callback` in the constructor and have a no-arg `listen()` reuse it — but functionally it would still only run during `listen()`.

valkey_glide_monitor.stub.php
prateek-kumar-improving: Other clients don't expose a dropped-lines method — they drop overflow silently. PHP's `getDroppedCount()` is an intentional, PHP-specific improvement: same bounded-queue strategy (bound memory so a slow consumer can't OOM the process), but instead of losing lines invisibly, PHP tells you how many were dropped so you can detect an incomplete capture.

This is the same mechanism discussed in the [`getDroppedCount()` thread](https://github.com/valkey-io/valkey-glide-php/pull/309#discussion_r3822826045), and the "count instead of silently drop" [approach](https://github.com/valkey-io/valkey-glide-php/pull/309#discussion_r3799818674).

valkey_glide_monitor.stub.php
prateek-kumar-improving: Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/94c268b

valkey_glide_monitor.stub.php
prateek-kumar-improving: Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/94c268b

valkey_glide_monitor.stub.php
prateek-kumar-improving: Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/94c268b

valkey_glide_monitor.stub.php
prateek-kumar-improving: Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/94c268b

valkey_glide_monitor_common.h
prateek-kumar-improving: In practice it's always populated — real MONITOR output always carries a command name, so the C# assumption holds. The `NULL` case is purely defensive: `command` is only left `NULL` if the FFI ever delivers a zero-length command (which shouldn't happen for genuine MONITOR traffic) or if the allocation fails under memory pressure. We keep the guard because a stray null-deref on those edge cases would be worse than the consumer simply rendering an empty command string.

PHP uses C code and therefore extra null checks are better in these cases as compared to other languages.

valkey_glide_monitor_common.h
prateek-kumar-improving: The bounded-queue-with-drop strategy matches the other clients — e.g. Go uses a fixed-size buffered channel and drops on overflow via `select { … default: }`. The difference is what happens on a drop: Go/Python discard silently, whereas PHP records the loss and surfaces it through `getDroppedCount()`, so a caller can tell whether their capture was complete. The `10000` depth is just a tuning choice (buffer more before dropping). This is the same mechanism discussed in the [`getDroppedCount()` thread](https://github.com/valkey-io/valkey-glide-php/pull/309#discussion_r3822826045), and the "count instead of silently drop" [approach](https://github.com/valkey-io/valkey-glide-php/pull/309#discussion_r3799818674).

valkey_glide_monitor_common.c
prateek-kumar-improving: Every client receives the same thing. The shared FFI (`create_monitor_client`) serializes args to a JSON string (`serde_json::to_string(&line.args)`) and passes `args_json`/`args_json_len` to all languages; the Go binding, for example, decodes it with `json.Unmarshal`. 

C# ends up with a string array only after its runtime JSON-decodes that same buffer. PHP was doing the same thing, just with a hand-rolled C decoder because `json_decode` can't be called from the background FFI thread. I've reworked it so the background thread only copies the raw JSON bytes, and the args are decoded with PHP's native `php_json_decode` on the main thread when the `ValkeyGlideMonitorLine` is built. Same result as the other clients, no custom parser.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/5901ae4

valkey_glide_monitor_common.c
prateek-kumar-improving: The JSON string decoder (escapes, `\uXXXX`, surrogate pairs, UTF-8 re-encoding) is removed. It only existed because we were decoding on the background FFI thread, where Zend APIs (and thus `json_decode`) aren't available. I've moved decoding to the PHP main thread and now use `php_json_decode`, so the entire `decode_json_string`/`parse_json_string_array` block (~190 lines) has been deleted.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/5901ae4

valkey_glide_monitor_common.c
prateek-kumar-improving: The custom parser this refers to has been removed entirely — args are now decoded with PHP's native `php_json_decode` on the main thread rather than a hand-rolled C parser, so there's no longer bespoke parsing logic to unit-test (we rely on PHP's well-tested JSON implementation, with a fallback to an empty array on malformed input). The existing integration tests still verify args are captured correctly.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/5901ae4

valkey_glide_monitor_common.c
prateek-kumar-improving: Refactored to remove the `goto`. I extracted the message construction into `build_monitor_message()` and the enqueue logic into `try_enqueue_monitor_message()` (with a small `dup_len_str()` helper), so the callback is now linear: look up the registry entry, build the message if the monitor is active, try to enqueue it, and free it inline if it was rejected — with a single `mutex_unlock` at the end and no labels after `return`. Behavior is unchanged.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/92fe08f

valkey_glide_monitor_common.c
prateek-kumar-improving: The slicing wasn't actually buying anything (the loop never checked for an interruption between slices, and PHP can't interrupt a blocking `cond_timedwait` mid-call anyway), and it could even under-wait slightly on spurious wakeups. I've replaced it with a single monotonic **deadline-based** wait.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/92fe08f

valkey_glide_pubsub_common.h
prateek-kumar-improving: Good point — these are generic threading primitives, not pub/sub-specific, and MONITOR only pulls in this header to get a mutex. I've taken you up on deferring it: filed #318 to extract the `mutex_*`/`cond_*` wrappers into a dedicated `valkey_glide_threading.{c,h}` module, and added a `TODO(#318)` here so it's tracked in-code.

Follow-up issue: https://github.com/valkey-io/valkey-glide-php/issues/318
Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/e4e4cb9

valkey_glide_pubsub_common.c
prateek-kumar-improving: Agreed, silently turning a bounded wait into an infinite one is a bad failure mode (and it was even returning success). `clock_gettime(CLOCK_REALTIME)` effectively never fails on a supported platform, but I've changed the fallback to return a timeout indication (non-zero) instead of doing an unbounded `pthread_cond_wait`. That matches the documented return contract, so a bounded caller re-checks its own predicate/deadline rather than hanging.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/e4e4cb9

valkey_glide_pubsub_common.c
prateek-kumar-improving: `pthread_cond_timedwait` requires `tv_nsec` in `[0, 1e9)`, and adding the timeout's sub-second part can overflow one second. Concretely: a current `tv_nsec` of 800,000,000 plus a 300 ms timeout (300,000,000 ns) = 1,100,000,000 ns, so we carry one second and keep 100,000,000 ns — otherwise the timespec is invalid and the call returns `EINVAL`. It happens whenever `current_ns + (timeout_ms % 1000)*1e6 >= 1e9`. I've added that example as a comment.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/e4e4cb9

valkey_glide.c
prateek-kumar-improving: Done — removed the redundant comment (the `#include` is self-explanatory).

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/1008f25

tests/ValkeyGlideMonitorTest.php
prateek-kumar-improving: Fixed — removed the extra blank line.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/1008f25

.gitignore
prateek-kumar-improving: Yes — these are generated at build time by PHP's `gen_stub.php` from `valkey_glide_monitor.stub.php` (via the gen_stub rule in `Makefile.frag`), so they're build artifacts rather than source. They're ignored for the same reason as the other `*_arginfo.h` entries just above them (`valkey_glide_arginfo.h`, `valkey_glide_cluster_arginfo.h`, `logger_arginfo.h`, …).

tests/ValkeyGlideMonitorTest.php
prateek-kumar-improving: Good points. I replaced the fixed 20-iteration loop with a deadline-based `while` loop and added a comment explaining why the loop exists: MONITOR streams every command the server processes — including traffic from other clients/tests on the shared server — so we scan until we find our own SET or a timeout elapses, rather than after an arbitrary number of reads.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/c9af670

tests/ValkeyGlideMonitorTest.php
prateek-kumar-improving: Agreed — since we issue the command ourselves, the assertions are now exact: `timestamp` >= a reference captured just before the SET (and not in the future), `db === 0`, `command === "set"` (MONITOR lowercases it), `args === [key, value]` in order, and the full `__toString` MONITOR prefix (`"<ts> [0 <addr>] \"set\""` plus key/value). The one field I left as a shape check rather than an exact match is `clientAddr` — it's the action client's ephemeral `host:port`, and the port is OS-assigned, so I assert the `host:port` format instead of a fixed value.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/c9af670

tests/ValkeyGlideMonitorTest.php
prateek-kumar-improving: Done — replaced the counted loop with a `while`-until-empty drain bounded by a deadline, then measure that a single `tryGetMonitorMessage()` returns promptly (which is the actual non-blocking behavior under test).

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/c9af670

tests/ValkeyGlideMonitorTest.php
prateek-kumar-improving: The two tests deliberately cover the **two different consumption APIs**:

- `testMonitorPullReturnsStructuredLine` exercises the **Queue/pull API** (`getMonitorMessage`/`tryGetMonitorMessage`), which the test process drives directly on the main thread.
- This test exercises the **Callback API** (`listen()`), which is a *blocking* loop that runs until the callback returns non-null. It can't run in the test-driver process because that same process must issue the commands being monitored — `listen()` would block it. So a child process runs `listen()` while the parent issues commands and coordinates via temp files, mirroring the existing pub/sub `subscribe()` tests.

So it isn't redundant coverage — the two tests hit different code paths. I've added a docblock on `testMonitorCapturesSetCommand` explaining this.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/373ac95

tests/scripts/monitor_listener.php
prateek-kumar-improving: This is a helper spawned as a **subprocess** by `testMonitorCapturesSetCommand` / `testMonitorCapturesGetCommand` (via `buildMonitorCommand()` + `proc_open`). It's needed because the callback API, `ValkeyGlideMonitor::listen()`, **blocks** until the callback returns non-null, so it can't run in the test-driver process (which must issue the commands being captured) — it runs in its own process instead, the same pattern the pub/sub tests use for the blocking `subscribe()`.

How it runs: `php monitor_listener.php <host> <port> <sync_file> <result_file> <expected_command> [max_lines]`. It opens a `ValkeyGlideMonitor`, runs `listen()`, writes `sync_file` on the **first** callback (a real readiness handshake — the parent sends PING probes meanwhile, proving the dedicated MONITOR connection is actually receiving events), and on seeing the expected command writes it to `result_file` and exits. This is documented in the header comment at the top of the script.

tests/scripts/monitor_multi_capture.php
prateek-kumar-improving: Same purpose and invocation as `monitor_listener.php`, used by `testMonitorCapturesMultipleCommands`: it runs the blocking `listen()` loop in a subprocess with the same sync-file readiness handshake. The difference is it **collects multiple** lines matching a prefix (up to `expected_count`) and writes them newline-separated to `result_file`, so the parent can assert several commands were all captured.

How it runs: `php monitor_multi_capture.php <host> <port> <sync_file> <result_file> <prefix> <expected_count>`. Behavior is documented in the script's header comment.

valkey_glide_monitor.c
```diff
@@ -0,0 +1,465 @@
...
+        if (EG(vm_interrupt)) {
+#endif
+            break;
+        }
+
+        monitor_message* msg = valkey_glide_monitor_dequeue(obj->info, 1000);
+        if (!msg) {
+            /* Timeout with no message: re-check active state and loop. */
+            bool active;
+            mutex_lock(&obj->info->queue_mutex);
```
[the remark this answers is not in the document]
prateek-kumar-improving: Code fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/70935a8

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,409 @@
...
+        // Free the native registry entries
+        mutex_lock(&monitor_registry_mutex);
+        monitor_registry_entry* entry = monitor_registry_head;
+        while (entry) {
+            monitor_registry_entry* next = entry->next;
+            free(entry);
+            entry = next;
+        }
+        monitor_registry_head = NULL;
+        mutex_unlock(&monitor_registry_mutex);
```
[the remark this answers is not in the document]
prateek-kumar-improving: Calling `cleanup_monitor_callback_info` here would be a use-after-free: `info` is `emalloc`'d (its cleanup does `zval_ptr_dtor` + `efree`), but this runs at MSHUTDOWN after the request heap is gone. In practice the registry is already empty by then — each monitor object's free handler runs teardown at request shutdown, which unregisters the entry and frees `info`. This loop only reclaims the native entry nodes; added a comment noting that.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/70935a8

valkey_glide_monitor_common.c
```diff
@@ -0,0 +1,389 @@
...
+        if (!info->is_active) {
+            mutex_unlock(&info->queue_mutex);
+            goto free_msg;
+        }
+        if (info->queue_depth >= MONITOR_QUEUE_MAX_DEPTH) {
+            // Queue full — drop and record the loss (surfaced via
+            // getDroppedCount()), rather than growing memory unbounded.
+            info->dropped_count++;
+            mutex_unlock(&info->queue_mutex);
+            goto free_msg;
```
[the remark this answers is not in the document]
prateek-kumar-improving: By design. Bounded-drop matches the other clients (Go/Python drop silently on a full buffer), and we already surface the loss via `getDroppedCount()` — cumulative and readable during and after the session. Overloading `listen()`'s return or throwing on a drop would be surprising given drops are intentional back-pressure, so `getDroppedCount()` is the explicit completeness signal.

valkey_glide_monitor.c
```diff
@@ -0,0 +1,479 @@
...
+        zval_ptr_dtor(&line);
+
+        if (stop)
+            break;
+    }
+
+    if (EG(exception)) {
+        RETURN_THROWS();
+    }
+    RETURN_TRUE;
```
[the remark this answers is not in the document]
prateek-kumar-improving: By design. Bounded-drop matches the other clients (Go/Python drop silently on a full buffer), and we already surface the loss via `getDroppedCount()` — cumulative and readable during and after the session. Overloading `listen()`'s return or throwing on a drop would be surprising given drops are intentional back-pressure, so `getDroppedCount()` is the explicit completeness signal.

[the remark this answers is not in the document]
prateek-kumar-improving: By design. Bounded-drop matches the other clients (Go/Python drop silently on a full buffer), and we already surface the loss via `getDroppedCount()` — cumulative and readable during and after the session. Overloading `listen()`'s return or throwing on a drop would be surprising given drops are intentional back-pressure, so `getDroppedCount()` is the explicit completeness signal.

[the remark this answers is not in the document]
prateek-kumar-improving: By design. Bounded-drop matches the other clients (Go/Python drop silently on a full buffer), and we already surface the loss via `getDroppedCount()` — cumulative and readable during and after the session. Overloading `listen()`'s return or throwing on a drop would be surprising given drops are intentional back-pressure, so `getDroppedCount()` is the explicit completeness signal.

.gitignore
currantw: Thanks for clarifying.

valkey_glide_monitor.stub.php
currantw: > // __toString renders the Valkey/Redis MONITOR line format.

1. Do other GLIDE clients have this same method?
2. I don't think we should generally mentioned Redis, since this is a _Valkey_ client (even if it is still compatible with some older versions of Redis). 🤷

tests/scripts/monitor_listener.php
currantw: Okay, sounds good!

valkey_glide_monitor.stub.php
currantw: > If cross-client familiarity matters in this case, we could also accept an optional $callback in the constructor and have a no-arg listen() reuse it — but functionally it would still only run during listen().

Why can't we just match the other clients: a callback in the constructor and no `listen` method at all?

valkey_glide_monitor.stub.php
currantw: Okay, I agree that this is an improvement. Can you raise GitHub issue(s) to add this for the other clients (except for C#, which doesn't have a queue)? Not sure if Ruby supports Monitor yet, so if it doesn't can you raise/update an issue to add support there, and add a comment to make sure the assignee does the same for Ruby?

valkey_glide_monitor_common.h
currantw: Hmmm. Okay, sounds good.

valkey_glide_monitor.stub.php
prateek-kumar-improving: This is mentioned above: **PHP cannot safely call user code off the main thread**. Will check more, but PHP does have languages specific limitations.

valkey_glide_pubsub_common.c
currantw: Here is a suggestion for simplifying this, which involves:

1. Creating constants instead of magic numbers
2. Convert everything to nanoseconds so we can avoid the separate `if` block:

```c
#define NSEC_PER_SEC  1000000000LL
#define NSEC_PER_MSEC 1000000LL

...

long long total_ns = (long long) ts.tv_nsec + (long long) timeout_ms * NSEC_PER_MSEC;
ts.tv_sec += total_ns / NSEC_PER_SEC;
ts.tv_nsec = total_ns % NSEC_PER_SEC;
```

currantw (approved):
Reviewed to 70935a8

valkey_glide_monitor.stub.php
prateek-kumar-improving: 1. No, they don't have it, was a PHP specific method. To keep parityremoved __toString. The tests/scripts that needed a text form now use a small helper method instead.
2. Also done — dropped the "Redis" wording.

Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/b02e9d4

valkey_glide_monitor.stub.php
prateek-kumar-improving: Thanks — filed https://github.com/valkey-io/valkey-glide/issues/6830 to add the dropped-line count to the other queueing clients (Go, Python, Java, Node), excluding C# since it has no queue. On Ruby: `valkey-glide-ruby` doesn't implement the MONITOR command yet (only a `MONITOR` request-type enum constant), so there's nothing to update there today — I've noted in the issue that when MONITOR is added to Ruby it should include the same dropped-count accounting, with an explicit call-out for the assignee.

valkey_glide_pubsub_common.c
prateek-kumar-improving: Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/85edaba

valkey_glide_monitor.stub.php
prateek-kumar-improving: It's a PHP limitation, not a preference. Other clients can run a constructor callback for the client's lifetime because their runtime invokes it on a background thread (Go goroutine, Python via the GIL, Node libuv, 
Java JNI). 

PHP's Zend engine is single-threaded and not thread-safe, with no GIL-equivalent — our MONITOR events arrive on a background Rust thread that can only do native C work and must never call into PHP. So a PHP callback can only run on the main thread during an active call; a constructor/lifetime callback would have nothing to drive it. 

So, listen() is that driver — a blocking loop that invokes the callback per line — exactly like PHPRedis subscribe()/psubscribe() and our own pub/sub. The pull API (getMonitorMessage/tryGetMonitorMessage) covers the non-blocking case. 

I can add an optional constructor $callback that a no-arg listen() reuses if you'd like the familiarity, but it would still only run during listen().

currantw: Thanks for the explanation 💪

valkey_glide_monitor.stub.php
currantw: Lovely, thank you

currantw (approved):
Looks good ✅

PR #314, opened 2026-08-17
feat(php): Implement lolwut command

### LINKED ISSUE #308: Implement LOLWUT command
## Description

Implement the `LOLWUT` command for PHP. Already available in Java, Python, Node, Go, and C#.

## Reference

- https://valkey.io/commands/lolwut/
- valkey-io/valkey-glide#6292 — Commands Tracker

prateek-kumar-improving:
### Summary

Implement the `LOLWUT` command for the PHP Valkey GLIDE extension, bringing the PHP client in line with the other GLIDE language clients.

### Issue link

This Pull Request is linked to issue: [Implement LOLWUT command](https://github.com/valkey-io/valkey-glide-php/issues/308)

Closes #308

- Adds `ValkeyGlide::lolwut(?int $version = null, ?array $parameters = null)`.
- Adds `ValkeyGlideCluster::lolwut(?int $version = null, ?array $parameters = null, mixed $route = null)`.
- Supports bare `LOLWUT`, `LOLWUT VERSION <version>`, and optional integer parameters.
- Returns a string for single-node execution and a node-address-keyed array for multi-node cluster routes.

### Implementation

- Registers the command in the standalone and cluster Zend APIs and adds PHP stub documentation.
- Marshals the optional `VERSION` token and numeric parameters through the core command argument framework using the existing `Lolwut` request type.
- Validates that the optional version and every supplied parameter are integers; command-specific arity and value validation remain server-side, matching the other GLIDE clients.
- Uses string response processing for standalone requests and cluster-aware string response processing for routed cluster requests.

### Limitations

- LOLWUT is not supported in cluster batch mode because the current PHP batch mechanism cannot preserve a per-command cluster route.
- The meaning and supported count of numeric parameters depend on the LOLWUT version and the Valkey server version.

### Testing

- Added standalone coverage for bare, empty, parameter-only, version-only, version-with-parameters, and Valkey 9 parameter forms, plus standalone transaction batching.
- Added cluster coverage for default, random-node, all-node, and Valkey 9 routing scenarios.
- Built successfully with `make` and verified both public method signatures by loading the compiled extension and inspecting them with reflection.
- `make test` could not execute integration tests in this environment because `valkey-server` is not installed.

### Checklist

- [x] Destination branch is correct - main or release.

### DIFF
--- a/tests/ValkeyGlideTest.php
+++ b/tests/ValkeyGlideTest.php
@@ -6666,7 +6666,23 @@ class ValkeyGlideTest extends ValkeyGlideBaseTest
     {
         $this->assertIsString($response);
         if (is_string($response)) {
-            $this->assertStringContains('ver', strtolower($response));
+            $response = strtolower($response);
+            $this->assertStringContains('ver', $response);
+            $this->assertStringContains(strtolower($this->version), $response);
+        }
+    }
+
+    protected function assertLolwutClusterResponses($responses): void
+    {
+        $this->assertIsArray($responses);
+        if (!is_array($responses)) {
+            return;
+        }
+
+        $this->assertGT(0, count($responses));
+        foreach ($responses as $node_address => $response) {
+            $this->assertIsString($node_address);
+            $this->assertLolwutResponse($response);
         }
     }
 
@@ -6683,6 +6699,22 @@ class ValkeyGlideTest extends ValkeyGlideBaseTest
             $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [30, 4]));
             $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [40, 20, 1, 2]));
         }
+    }
+
+    public function testLolwutInvalidArguments()
+    {
+        $this->assertThrows(
+            ValkeyGlideException::class,
+            fn() => $this->valkey_glide->lolwut('6')
+        );
+        $this->assertThrows(
+            ValkeyGlideException::class,
+            fn() => $this->valkey_glide->lolwut(null, '50,20')
+        );
+        $this->assertThrows(
+            ValkeyGlideException::class,
+            fn() => $this->valkey_glide->lolwut(null, [50, '20'])
+        );
     }
 
     public function testLolwutBatch()
--- a/tests/ValkeyGlideTest.php
+++ b/tests/ValkeyGlideTest.php
@@ -6662,6 +6662,43 @@ class ValkeyGlideTest extends ValkeyGlideBaseTest
 
 
 
+    protected function assertLolwutResponse($response): void
+    {
+        $this->assertIsString($response);
+        if (is_string($response)) {
+            $this->assertStringContains('ver', strtolower($response));
+        }
+    }
+
+    public function testLolwut()
+    {
+        $this->assertLolwutResponse($this->valkey_glide->lolwut());
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(null, []));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [50, 20]));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(5));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(5, [30, 4, 4]));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(6, [50, 20]));
+
+        if ($this->is_valkey && $this->minVersionCheck('9.0.0')) {
+            $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [30, 4]));
+            $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [40, 20, 1, 2]));
+        }
+    }
+
+    public function testLolwutBatch()
+    {
+        $responses = $this->valkey_glide->multi()
+            ->lolwut()
+            ->lolwut(5)
+            ->lolwut(6, [50, 20])
+            ->exec();
+
+        $this->assertIsArray($responses, 3);
+        foreach ($responses as $response) {
+            $this->assertLolwutResponse($response);
+        }
+    }
+
     /**
      * Scan and variants
      */
--- a/tests/ValkeyGlideTest.php
+++ b/tests/ValkeyGlideTest.php
@@ -6695,22 +6695,48 @@ class ValkeyGlideTest extends ValkeyGlideBaseTest
         $this->assertLolwutResponse($this->valkey_glide->lolwut(5, [30, 4, 4]));
         $this->assertLolwutResponse($this->valkey_glide->lolwut(6, [50, 20]));
 
+        // Scalar arguments are coerced like other integer arguments.
+        $this->assertLolwutResponse($this->valkey_glide->lolwut('6'));
+
+        // Parameters holding references are dereferenced before sending.
+        $width = 50;
+        $height = 20;
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [&$width, &$height]));
+
         if ($this->is_valkey && $this->minVersionCheck('9.0.0')) {
             $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [30, 4]));
             $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [40, 20, 1, 2]));
         }
     }
 
+    public function testLolwutParametersAffectOutput()
+    {
+        // The response size scales with the requested canvas, so a larger
+        // canvas must produce a longer reply. This proves the numeric
+        // parameters actually reach the server rather than being dropped.
+        $small = $this->valkey_glide->lolwut(null, [10, 2]);
+        $large = $this->valkey_glide->lolwut(null, [100, 20]);
+
+        $this->assertIsString($small);
+        $this->assertIsString($large);
+        if (is_string($small) && is_string($large)) {
+            $this->assertGT(strlen($small), strlen($large));
+        }
+    }
+
     public function testLolwutInvalidArguments()
     {
+        // A non-numeric version cannot be coerced to int -> TypeError.
         $this->assertThrows(
-            ValkeyGlideException::class,
-            fn() => $this->valkey_glide->lolwut('6')
+            TypeError::class,
+            fn() => $this->valkey_glide->lolwut('abc')
         );
+        // Parameters must be an array -> TypeError.
         $this->assertThrows(
-            ValkeyGlideException::class,
+            TypeError::class,
             fn() => $this->valkey_glide->lolwut(null, '50,20')
         );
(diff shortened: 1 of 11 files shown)


valkey_glide_commands.c
```diff
@@ -1339,6 +1339,110 @@ int execute_time_command(zval* object, int argc, zval* return_value, zend_class_
...
+    if (version && Z_TYPE_P(version) != IS_NULL && Z_TYPE_P(version) != IS_LONG) {
+        zend_throw_exception(
+            get_valkey_glide_exception_ce(), "LOLWUT version must be an integer", 0);
+        return 0;
+    }
+    if (parameters && Z_TYPE_P(parameters) != IS_NULL && Z_TYPE_P(parameters) != IS_ARRAY) {
+        zend_throw_exception(
+            get_valkey_glide_exception_ce(), "LOLWUT parameters must be an array of integers", 0);
+        return 0;
+    }
```
Aryex: Potentially dead code. This guard won't be hit because `parameter` and `version` are statically typed. The zend engine would type check them and return a TypeError. 

Could you verify this is the case?

Aryex (approved):

valkey_glide_commands.c
xShinnRyuu, line 1386: The declared signature and the runtime contract disagree about what these two arguments accept.

The stubs declare `?int $version` and `?array $parameters`, `gen_stub.php` turns that into `IS_LONG` / `IS_ARRAY` arginfo, and Reflection reports `?int` / `?array`. Argument types in arginfo are not enforced for internal functions, so `z!` accepts anything and the two manual checks below are what actually decide. Values the declared signature permits get rejected:

```php
$client->lolwut('6');         // ValkeyGlideException: LOLWUT version must be an integer
$client->lolwut(6.0);         // same
$client->expire('k', '100');  // true, PHP coerced the string
```

`"O|z!z!"` and `"O|z!z!z!"` are the only ZPP specs in the extension backing stub-declared scalar types. Elsewhere a stub `int` is parsed with `l` and a stub `array` with `a`, which is why `expire` coerces and `lolwut` does not.

Two ways to line them up. Parse with `"O|l!a!"` (`"O|l!a!z!"` for cluster) and drop both checks: I built that variant, and `lolwut('6')` and `lolwut(6.0)` then work while `lolwut('abc')` raises `ValkeyGlide::lolwut(): Argument #1 ($version) must be of type ?int, string given`. That does move the two top-level cases in `testLolwutInvalidArguments` from `ValkeyGlideException` to `TypeError`, and `TestSuite::assertThrows` catches `Exception` only, so it would need a `Throwable` catch. Otherwise keep the checks and widen the stubs to `mixed $version` and `mixed $parameters`, so the reflected signature matches what the method really takes.

valkey_glide_commands.c
```diff
@@ -1339,6 +1339,110 @@ int execute_time_command(zval* object, int argc, zval* return_value, zend_class_
...
+
+    valkey_glide = VALKEY_GLIDE_PHP_ZVAL_GET_OBJECT(valkey_glide_object, object);
+    if (!valkey_glide || !valkey_glide->glide_client) {
+        return 0;
+    }
+
+    /* Cluster batches cannot preserve a per-command route. */
+    if (is_cluster && valkey_glide->is_in_batch_mode) {
+        return 0;
+    }
```
xShinnRyuu: This rejects every cluster batch call, including the ones with no route to preserve. The check runs before argument parsing, so `$cluster->multi()->lolwut(6)->exec()` gets `false` back from `lolwut()` and the chain breaks on the next call even though no route was supplied.

Cluster batches handle routeless commands fine: `set` and `get` queue and `exec` normally on a cluster client. Moving this below `zend_parse_method_parameters` and gating it on `route && Z_TYPE_P(route) != IS_NULL` recovers the routeless case. I built that variant against a 3-shard cluster and `$cluster->multi()->lolwut()->lolwut(6, [10, 2])->exec()` returned both strings, while `lolwut(null, null, 'allNodes')` was still rejected.

A second consequence of checking before parsing is that the version and parameter validation never runs in cluster batch mode, so `$cluster->multi(); $cluster->lolwut('nonsense');` returns `false` instead of throwing.

valkey_glide_commands.c
```diff
@@ -1339,6 +1339,110 @@ int execute_time_command(zval* object, int argc, zval* return_value, zend_class_
...
+
+        core_args.all_args[arg_index].type                = CORE_ARG_TYPE_LONG;
+        core_args.all_args[arg_index].data.long_arg.value = Z_LVAL_P(version);
+        arg_index++;
+    }
+
+    if (parameters && Z_TYPE_P(parameters) == IS_ARRAY) {
+        zval* parameter;
+        ZEND_HASH_FOREACH_VAL(Z_ARRVAL_P(parameters), parameter) {
+            if (Z_TYPE_P(parameter) != IS_LONG) {
```
xShinnRyuu, line 1428: An element holding a reference fails this check, because `ZEND_HASH_FOREACH_VAL` yields the `IS_REFERENCE` wrapper rather than the value:

```php
$w = 10; $h = 2;
$client->lolwut(null, [&$w, &$h]);  // ValkeyGlideException: LOLWUT parameters must contain only integers
```

The extension's other array arguments unwrap references because they pass through helpers that do it for them: `prepare_multi_key_args` uses `convert_to_string` and `prepare_key_value_pairs_args` uses `zval_get_string`. Both `$client->del([&$key])` and `$client->mSet(['k' => &$v])` work today. Adding `ZVAL_DEREF(parameter);` as the first line of the loop body brings this in line with them, and I confirmed the by-reference call then succeeds.

tests/ValkeyGlideTest.php
```diff
@@ -6662,6 +6662,75 @@ public function testTime()
...
+    }
+
+    public function testLolwut()
+    {
+        $this->assertLolwutResponse($this->valkey_glide->lolwut());
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(null, []));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(null, [50, 20]));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(5));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(5, [30, 4, 4]));
+        $this->assertLolwutResponse($this->valkey_glide->lolwut(6, [50, 20]));
```
xShinnRyuu, line 6714: These calls all funnel into `assertLolwutResponse`, which checks only that the reply is a string containing `ver` and the server version. A bare `LOLWUT` satisfies that, so nothing here would notice `$parameters` never reaching the server. I sent a bare `LOLWUT` through `rawCommand` against Valkey 9.0.1 and got back a 3345 byte string containing both `ver` and `9.0.1`.

The `VERSION` token itself is covered by accident: dropping it turns `lolwut(5)` into `LOLWUT 5`, and putting the parameters first turns `lolwut(6, [50, 20])` into `LOLWUT 50 20 VERSION 6`. Valkey 9 rejects both with a syntax error, so `assertIsString` catches those. The parameters are the gap.

Reply size tracks the requested canvas, so one relative assertion closes it. Measured through the built extension over ten runs, `lolwut(null, [10, 2])` came back at 127 to 129 bytes and `lolwut(null, [100, 20])` at 2125 to 2127. Comparing the two lengths avoids depending on the art, which shifts by a byte or two per call.

valkey_glide_commands.c
prateek-kumar-improving: Fixed. Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/7038c29a93cfa1e2f4c60e671d66883fe13cbe13

valkey_glide_commands.c
prateek-kumar-improving: Fixed. Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/7038c29a93cfa1e2f4c60e671d66883fe13cbe13

valkey_glide_commands.c
prateek-kumar-improving: Fixed. Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/7038c29a93cfa1e2f4c60e671d66883fe13cbe13

tests/ValkeyGlideTest.php
prateek-kumar-improving: Fixed. Code updated in this commit: https://github.com/valkey-io/valkey-glide-php/commit/7038c29a93cfa1e2f4c60e671d66883fe13cbe13
Added testLolwutParametersAffectOutput()

valkey_glide_commands.c
xShinnRyuu: Confirmed at 7038c29a9. `valkey_glide_commands.c:1362` and `:1373` now parse with `"O|l!a!z!"` and `"O|l!a!"`, and both manual type checks are gone. The generated arginfo still declares `?int $version` / `?array $parameters`, so the declared signature and the runtime now agree.

Built the extension at this head and ran it against Valkey 9.0.1: `lolwut('6')` returns a string, `lolwut(6.0)` returns a string, and `lolwut('abc')` raises `TypeError: ValkeyGlide::lolwut(): Argument #1 ($version) must be of type ?int, string given`. `lolwut(null, '50,20')` raises the matching TypeError for `$parameters`.

`tests/TestSuite.php:844` widened `assertThrows` from `Exception` to `Throwable`, and `testLolwutInvalidArguments` now expects `TypeError` for both top-level cases. All four lolwut tests pass in `ValkeyGlideTest`.

valkey_glide_commands.c
xShinnRyuu: Confirmed at 7038c29a9. `valkey_glide_commands.c:1427` adds `ZVAL_DEREF(parameter);` as the first statement in the loop body.

Built and ran it against Valkey 9.0.1: `$client->lolwut(null, [&$w, &$h])` with `$w = 50, $h = 20` returns a 1127 byte string instead of throwing. Same result on a cluster client. `tests/ValkeyGlideTest.php:6701-6704` covers it.

tests/ValkeyGlideTest.php
xShinnRyuu: Confirmed at 7038c29a9. `tests/ValkeyGlideTest.php:6712-6725` adds `testLolwutParametersAffectOutput`, which compares the reply length for `lolwut(null, [10, 2])` against `lolwut(null, [100, 20])`.

Measured on the build at this head against Valkey 9.0.1 over three runs: 127 to 128 bytes against 2127 bytes. A bare `LOLWUT` is 3345 bytes, so a dropped parameter list no longer passes. `testLolwut` also adds `lolwut('6')` and a by-reference parameter case at lines 6698-6704. The tests pass in both the standalone and cluster suites.

xShinnRyuu (approved):

PR #315, opened 2026-08-18
feat(php): implement LASTSAVE command

### LINKED ISSUE #307: Implement LASTSAVE command
## Description

Implement the `LASTSAVE` command for PHP. Already available in Java, Python, Node, Go, and C#.

## Reference

- https://valkey.io/commands/lastsave/
- valkey-io/valkey-glide#6292 — Commands Tracker

prateek-kumar-improving:
### Summary

Implement the `LASTSAVE` command for the PHP Valkey GLIDE extension, bringing the PHP client in line with the other GLIDE language clients.

### Issue link

This Pull Request is linked to issue: [Implement LASTSAVE command](https://github.com/valkey-io/valkey-glide-php/issues/307)

Closes #307

- Adds `ValkeyGlide::lastSave()` returning an integer Unix timestamp.
- Adds `ValkeyGlideCluster::lastSave(mixed $route = null)` with optional cluster routing.
- For multi-node routes (`allNodes`, `allPrimaries`), returns an associative array mapping node addresses to integer timestamps.
- Introduces a reusable `process_core_cluster_int_result()` processor for cluster commands that return integers with multi-node route support.

### Implementation

- Registers the command in both standalone and cluster Zend APIs via `LASTSAVE_METHOD_IMPL` macro.
- Uses zero-argument serialization (`prepare_zero_args`) with the `LastSave` protobuf request type.
- Cluster routing follows the established optional-route pattern: parses variadic args, sets `has_route`/`route_param` when a non-null route is provided.
- Created `process_core_cluster_int_result` which wraps `process_core_cluster_result` with `process_core_int_result` as the per-node handler. This ensures Map responses from multi-node routes produce a PHP associative array, while single-node routes return a plain integer.

### Limitations

None. LASTSAVE works in both standalone and cluster batch/pipeline mode (unrouted).

### Testing

- Standalone: default invocation asserts timestamp > yesterday; transaction batch asserts the same.
- Cluster: covers default (no route), explicit null route, `randomNode`, `allNodes`, and `allPrimaries`. Multi-node tests assert non-empty associative arrays with string keys and integer values > yesterday.
- Cluster pipeline batch: tests unrouted LASTSAVE in pipeline mode.
- Built successfully with `make` (`-Werror`); C code passes `clang-format-18`.
- Extension API verified via reflection.
- Integration tests require `valkey-server` which is not available in this environment.

### Checklist

- [x] Destination branch is correct - main or release.

### DIFF
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -1422,6 +1422,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertGT(0, count($result));
         foreach ($result as $node_address => $timestamp) {
             $this->assertIsString($node_address);
+            $this->assertStringContains(':', $node_address);
             $this->assertIsInt($timestamp);
             $this->assertGT(0, $timestamp);
         }
@@ -1432,6 +1433,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertGT(0, count($result));
         foreach ($result as $node_address => $timestamp) {
             $this->assertIsString($node_address);
+            $this->assertStringContains(':', $node_address);
             $this->assertIsInt($timestamp);
             $this->assertGT(0, $timestamp);
         }
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -1401,7 +1401,7 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testLastSave()
     {
-        $yesterday = time() - 86400;
+        $yesterday = time() - self::SECONDS_PER_DAY;
 
         // Default (random node)
         $result = $this->valkey_glide->lastSave();
@@ -1436,6 +1436,17 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
 
     public function testLastSaveBatch()
     {
+        if (!$this->havePipeline()) {
+            $this->markTestSkipped('Pipeline not supported');
+            return;
+        }
+
+        $yesterday = time() - self::SECONDS_PER_DAY;
+        $this->valkey_glide->pipeline();
+        $this->valkey_glide->lastSave();
+        $result = $this->valkey_glide->exec();
+        $this->assertIsArray($result);
+        $this->assertGT($yesterday, $result[0]);
     }
 
     public function testScan()
--- a/tests/ValkeyGlideClusterTest.php
+++ b/tests/ValkeyGlideClusterTest.php
@@ -1399,6 +1399,45 @@ class ValkeyGlideClusterTest extends ValkeyGlideTest
         $this->assertEquals(strval(intval($usec)), strval($usec));
     }
 
+    public function testLastSave()
+    {
+        $yesterday = time() - 86400;
+
+        // Default (random node)
+        $result = $this->valkey_glide->lastSave();
+        $this->assertGT($yesterday, $result);
+
+        // Explicit null route
+        $result = $this->valkey_glide->lastSave(null);
+        $this->assertGT($yesterday, $result);
+
+        // randomNode route
+        $result = $this->valkey_glide->lastSave('randomNode');
+        $this->assertGT($yesterday, $result);
+
+        // allNodes route
+        $result = $this->valkey_glide->lastSave('allNodes');
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
+        foreach ($result as $node_address => $timestamp) {
+            $this->assertIsString($node_address);
+            $this->assertGT($yesterday, $timestamp);
+        }
+
+        // allPrimaries route
+        $result = $this->valkey_glide->lastSave('allPrimaries');
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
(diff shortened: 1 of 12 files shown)


valkey_glide_core_common.c
```diff
@@ -1725,6 +1726,16 @@ int process_core_status_string_result(CommandResponse* response, void* output, z
     return process_core_cluster_result(response, output, return_value, process_core_string_result);
 }
 
+/**
+ * Cluster-aware integer result processor.
+ * For single-node: returns the integer value.
+ * For multi-node routes: returns an associative array of node => int.
+ * Used by commands like LASTSAVE and DBSIZE.
```
Aryex: Can you verify this claim? It was flagged as incorrect.

Aryex (approved):

valkey_glide_core_common.c
xShinnRyuu: The DBSIZE half of this line does not hold. `process_core_cluster_int_result` has exactly one caller, `execute_lastsave_command`. DBSIZE passes `process_core_int_result` directly (`execute_dbsize_command` in `valkey_glide_commands_3.c`), and it has to: glide-core gives DBSIZE `ResponsePolicy::Aggregate(AggregateOp::Sum)`, so a multi-node DBSIZE comes back as one summed integer, never a per-node map.

Measured on a 3-primary/3-replica cluster: `dbSize('allNodes')` returns an `int`, while `lastSave('allNodes')` returns a 6-entry array keyed by node address. So the single-node and multi-node lines above are accurate, only this last one is wrong.

Suggest replacing the DBSIZE reference with the condition that actually selects this processor, for example:

```c
/**
 * Cluster-aware integer result processor.
 * For single-node: returns the integer value.
 * For multi-node routes: returns an associative array of node => int.
 * Applies to integer-returning commands that have no glide-core response
 * policy, so a multi-node route arrives as a per-node Map. Currently LASTSAVE.
 */
```

tests/ValkeyGlideClusterTest.php
```diff
@@ -1399,6 +1399,59 @@ public function testTime()
...
+        $result = $this->valkey_glide->lastSave('randomNode');
+        $this->assertIsInt($result);
+        $this->assertGT(0, $result);
+
+        // allNodes route
+        $result = $this->valkey_glide->lastSave('allNodes');
+        $this->assertIsArray($result);
+        $this->assertGT(0, count($result));
+        foreach ($result as $node_address => $timestamp) {
+            $this->assertIsString($node_address);
```
xShinnRyuu, line 1424: Worth asserting the key shape here, not just that the key is a string. The helper this PR adds documents its multi-node result as "an associative array of node => int", and this loop is the only executable proof of the node half. `assertIsString` passes for any string key.

`testBgSave`, `testBgSaveSchedule` and `testBgRewriteAof` in this file go through the same `process_core_cluster_result` map branch, and each puts `$this->assertStringContains(':', $nodeAddress);` next to the `assertIsString`. Adding that line here and in the `allPrimaries` loop below would make the test confirm the documented contract. Keys came back as `127.0.0.1:7001` style addresses when I ran it.

valkey_glide_core_common.c
prateek-kumar-improving: Fixed. Code updated in commit: https://github.com/valkey-io/valkey-glide-php/commit/2a0a5008653e4d60f37fe7b383ece6bedbd91fce

prateek-kumar-improving: Fixed. Code updated in commit: https://github.com/valkey-io/valkey-glide-php/commit/2a0a5008653e4d60f37fe7b383ece6bedbd91fce

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: Fixed. Code updated in commit: https://github.com/valkey-io/valkey-glide-php/commit/2a0a5008653e4d60f37fe7b383ece6bedbd91fce

valkey_glide_core_common.c
xShinnRyuu: Confirmed at 2a0a500. The DBSIZE reference is gone from that doc comment. `valkey_glide_core_common.c:1733-1734` now reads:

```
 * Applies to integer-returning commands that have no glide-core response
 * policy, so a multi-node route arrives as a per-node Map. Currently LASTSAVE.
```

I re-checked the new wording against head: `process_core_cluster_int_result` still has one caller, `execute_lastsave_command` (`valkey_glide_commands.c:1380`), and `execute_dbsize_command` still passes `process_core_int_result` (`valkey_glide_commands_3.c:2111`). Resolving.

tests/ValkeyGlideClusterTest.php
xShinnRyuu: Confirmed at 2a0a500. `$this->assertStringContains(':', $node_address);` is now in both loops: `tests/ValkeyGlideClusterTest.php:1425` for `allNodes` and `:1436` for `allPrimaries`. That matches `testBgSave` (`:465`), `testBgSaveSchedule` (`:484`) and `testBgRewriteAof` (`:617`) in the same file, and the argument order is right for the helper at `tests/TestSuite.php:730` (needle first). Resolving.

xShinnRyuu (approved):

PR #320, opened 2026-08-24
feat(php): implement CONFIG command with routing for cluster

### LINKED ISSUE #284: Implement config command in cluster
Implement config command in cluster mode. For now we are using rawCommand in tests where we have dependency on config commands. Update the tests once this is implemented.

prateek-kumar-improving:
### Summary

Implements the `CONFIG` command for the `ValkeyGlideCluster` client with per-request routing. Cluster tests that previously depended on `rawCommand('...', 'CONFIG', ...)` now use the native `config()` method.

### Issue link

This Pull Request is linked to issue: [Implement config command in cluster](https://github.com/valkey-io/valkey-glide-php/issues/284)
Closes #284

- Adds `ValkeyGlideCluster::config(mixed $route, string $operation, array|string|null $key_or_settings = null, ?string $value = null): mixed`.
- Supports the `GET`, `SET`, `RESETSTAT`, and `REWRITE` operations (same set as the standalone client).
- The routing configuration is the first argument, consistent with the existing cluster methods (`dbSize`, `client`, `rawCommand`).
- Response shapes for `CONFIG GET`:
  - A single-node route returns a flat associative array of `parameter => value`.
  - A multi-node route returns a per-node associative array keyed by node address, where each value is that node's `parameter => value` map.
  - `SET`/`RESETSTAT`/`REWRITE` return `true` on success.

### Implementation

- `valkey_glide_cluster.stub.php`: replaced the `//TODO` stub with the real `config()` method signature and documentation; regenerated arginfo.
- `valkey_glide_commands_3.c` (`execute_config_command`): detect cluster mode and, when not in batch mode, parse the leading route argument and dispatch via `execute_command_with_route()`; standalone and cluster-batch paths keep the existing `execute_command()` behavior. The shared argument-building and response-processing logic is reused, so the existing recursive map conversion handles the nested multi-node `CONFIG GET` response with no changes.
- The C method macro `CONFIG_METHOD_IMPL(ValkeyGlideCluster)` was already present; this PR wires it up through the stub and adds routing support in the shared helper.

### Limitations

- Routing is explicit (the route argument is required as the first parameter), matching the existing cluster command pattern; the client does not apply per-subcommand default routes.
- In cluster batch mode, the `$route` argument is ignored: the buffered `config()` command executes against glide-core's default routing at `exec()` time (there is no per-command batch route), consistent with other cluster commands.

### Testing

- Implemented `ValkeyGlideClusterTest::testConfig()` (previously skipped), covering: `GET` single param, `GET` multiple params (7.0+), `GET` wildcard, `GET` with a multi-node route (per-node map), `GET` with no args (error), `SET` single param with verification, `SET` array of params (7.0+), routed `SET`, `RESETSTAT`, `REWRITE` (boolean), and malformed calls.
- Switched `triggerLatencySpike()` from `rawCommand` to `config()`.
- Verified against a live Valkey 9.1.1 cluster and a standalone server:
  - Cluster `testConfig` — PASSED
  - Cluster latency tests (10, now depending on `config()`) — all PASSED
  - Standalone `testConfig` (shared-function refactor) — PASSED

### Checklist


valkey_glide_commands_3.c
```diff
@@ -1475,21 +1475,58 @@ static int process_config_command_respose(CommandResponse* response,
...
+    } else {
+        /* Standalone mode (or cluster batch mode): parse operation and optional key/value */
+        if (zend_parse_method_parameters(
+                argc, object, "Os|z!z!", &object, ce, &operation, &operation_len, &key, &value) ==
+            FAILURE) {
+            return 0;
+        }
+        /* Re-fetch the object in case parsing updated it */
+        valkey_glide = VALKEY_GLIDE_PHP_ZVAL_GET_OBJECT(valkey_glide_object, object);
+    }
```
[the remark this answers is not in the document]
prateek-kumar-improving: Updated. Also, added a test called testConfigOperationsBatch in tests/ValkeyGlideClusterBatchTest.php

valkey_glide_commands_3.c
```diff
@@ -1478,18 +1478,60 @@ int execute_config_command(zval* object, int argc, zval* return_value, zend_clas
...
+        }
+        operation     = Z_STRVAL(z_args[1]);
+        operation_len = Z_STRLEN(z_args[1]);
+
+        if (z_argc >= 3 && Z_TYPE(z_args[2]) != IS_NULL) {
+            key = &z_args[2];
+        }
+        if (z_argc >= 4 && Z_TYPE(z_args[3]) != IS_NULL) {
+            value = &z_args[3];
+        }
```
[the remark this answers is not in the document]
prateek-kumar-improving: Updated in https://github.com/valkey-io/valkey-glide-php/commit/586606119fcdf157736983b4ff0a6fa50cbd94e3

valkey_glide_commands_3.c
```diff
@@ -1679,9 +1721,15 @@ int execute_config_command(zval* object, int argc, zval* return_value, zend_clas
...
+            if (route) {
+                result = execute_command_with_route(
+                    valkey_glide->glide_client, command_type, arg_count, args, args_len, route);
+            } else {
+                result = execute_command(
+                    valkey_glide->glide_client, command_type, arg_count, args, args_len);
+            }
 
             if (result) {
                 if (result->command_error) {
```
xShinnRyuu, line 1736: The PR description states this change "fixed a pre-existing leak on the command_error path (now goto cleanup)", but this command_error branch still does `return 0;` a few lines below, which bypasses the `cleanup:` label. `args` and `args_len` (and, for the array GET/SET forms, `temp_strings` and its estrdup'd entries) are allocated before this point and are not freed on this path. Base and head are byte-identical here, so the claimed fix is not present in the diff.

These are request-scoped emalloc allocations, so a debug build's leak detector will flag them and they accumulate within a request that issues repeatedly failing CONFIG commands (for example a routed `CONFIG SET` of an unknown parameter). Suggested fix: set `status = 0;` and `goto cleanup;` in place of the `return 0;` (result is already freed just above), or correct the description to match what the diff actually does.

valkey_glide_commands_3.c
```diff
@@ -1478,18 +1478,60 @@ int execute_config_command(zval* object, int argc, zval* return_value, zend_clas
...
+        }
+
+        /* Need at least the route and the operation, and at most route + operation + key + value */
+        if (z_argc < 2 || z_argc > 4) {
+            return 0;
+        }
+
+        /* Only route when not batching; in batch mode routing is applied to the whole batch */
+        if (!valkey_glide->is_in_batch_mode) {
+            route = &z_args[0];
```
xShinnRyuu, line 1509: In batch mode the route is intentionally dropped here and is never forwarded to buffer_command_for_batch. execute_exec_command parses "O" and accepts no route argument, and batch() is invoked with NULL options, so there is no batch-level routing path either. The net effect is that a routed call such as multi()->config('allNodes','SET',...)->exec() is buffered and then executed against glide-core's default routing, silently discarding the caller's explicit route.

This is consistent with the other cluster commands and is disclosed under Limitations, but the wording "batches are routed at the batch level" is inaccurate since no batch-level route is exposed. Consider either rejecting a non-default route in batch mode with a clear error, or rewording the limitation to state plainly that the route argument is ignored in batch mode.

valkey_glide_commands_3.c
```diff
@@ -1478,18 +1478,60 @@ int execute_config_command(zval* object, int argc, zval* return_value, zend_clas
...
+         * non-batch execution path below, since batches are routed at execution time. */
+        zval* z_args = NULL;
+        int   z_argc = 0;
+        if (zend_parse_method_parameters(argc, object, "O*", &object, ce, &z_args, &z_argc) ==
+            FAILURE) {
+            return 0;
+        }
+
+        /* Need at least the route and the operation, and at most route + operation + key + value */
+        if (z_argc < 2 || z_argc > 4) {
```
xShinnRyuu, line 1503: Minor: within this 2 to 4 argument range, RESETSTAT and REWRITE (and GET when a stray value is supplied) parse the extra trailing arguments into key/value and then silently ignore them rather than rejecting them. For example config('randomNode', 'REWRITE', 'x', 'y') is accepted. This matches the existing lenient behavior, so it is optional, but a per-operation arity check would make malformed calls fail loudly instead of being accepted.

tests/ValkeyGlideClusterTest.php
```diff
@@ -173,10 +173,100 @@ public function testReconnectSelect()
...
-    /* These 'directed node' commands work differently in ValkeyGlideCluster */
+    /* These 'directed node' commands work differently in ValkeyGlideCluster:
+       they take a routing configuration as the first argument. */
     public function testConfig()
     {
-        $this->markTestSkipped();
+        /* CONFIG GET against a single node returns a flat associative array */
+        $cfg = $this->valkey_glide->config('randomNode', 'GET', 'timeout');
+        $this->assertArrayKey($cfg, 'timeout');
+        $sec = $cfg['timeout'];
```
xShinnRyuu, line 183: Minor test-hygiene note: `$sec` is captured from a single random node, and the restore in the finally block writes it to all nodes via config('allNodes','SET','timeout',$sec). If the nodes ever started with divergent timeout values, this restore would normalize every node to the one sampled value rather than restoring the prior per-node state. The same pattern appears in the 7.0+ multi-parameter section below. Cluster nodes usually start with identical config so this is unlikely to bite in practice, but capturing per-node values (or asserting uniformity first) would make the round-trip exact.

valkey_glide_commands_3.c
prateek-kumar-improving: Fixed. Reworded the code comment and the `@return` doc to state that the `$route` argument is ignored in batch mode: it's consumed so the operation isn't mis-parsed, but it's not forwarded to the buffered command, which executes against glide-core's default routing at `exec()`. This matches the other cluster commands' batch behavior. Fixed in https://github.com/valkey-io/valkey-glide-php/commit/d915fb141fea5f428f21f035040486b1d1dd9225.

valkey_glide_commands_3.c
prateek-kumar-improving: The description was leftover from an earlier revision where we added `goto cleanup` on this path; that change was reverted but it was not removed from the corresponding line in the description. Corrected the description so it now matches the code (`return 0;` unchanged from base). The pre-existing error-path behavior is unchanged by this PR.

valkey_glide_commands_3.c
prateek-kumar-improving: The standalone `config()` already ignores extra trailing args, so I kept the cluster version the same for consistency. If we'd rather reject extra args, we should change both — happy to do that here if you want it.

tests/ValkeyGlideClusterTest.php
prateek-kumar-improving: It's a per-node-divergence edge case. The `try/finally` guarantees the sampled value is written back on any mid-test failure, and the cluster nodes start with uniform `timeout`/`set-max-intset-entries`. Left out per-node `routeByAddress` capture/restore to avoid the extra complexity for settings that are uniform in the test cluster.

valkey_glide_commands_3.c
xShinnRyuu: might need to update your PR description as well

xShinnRyuu (approved):

valkey_glide_commands_3.c
prateek-kumar-improving: Updated.

Aryex (approved):

PR #321, opened 2026-08-25
feat(config): add mTLS (mutual TLS) client certificate support

### LINKED ISSUE #288: feat(config): add mTLS (mutual TLS) support
### Describe the feature

Add mutual TLS (mTLS) support to the Valkey GLIDE PHP client. mTLS enables two-way certificate authentication between the client and the Valkey server, where both parties present and verify certificates. This is already supported in other GLIDE language clients (Java, Python, Node.js) but is currently missing from the PHP implementation.

### Use Case

Secure enterprise deployments often require mTLS to ensure that only authorized clients can connect to database servers. Without mTLS support, PHP users cannot:
* Meet zero-trust network security requirements that mandate mutual authentication.
* Deploy Valkey GLIDE PHP in environments where server-only TLS is insufficient (e.g., regulated industries, multi-tenant cloud environments).
* Achieve feature parity with other GLIDE language clients that already support mTLS.

### Proposed Solution

Extend the advanced_config['tls_config'] array to accept client_cert and client_key parameters for mTLS authentication. The implementation should:

1. Add client_cert and client_key fields to valkey_glide_tls_advanced_configuration_t in common.h.
2. Parse the new options from the tls_config array in the connection configuration (similar to how root_certs is handled today).
3. Pass the client credentials through to glide-core via the protobuf ConnectionRequest fields (client_cert = field 22, client_key = field 23) in valkey_glide_core_commands.c.
4. Support both file-path-based and in-memory (PEM string) certificate loading.
5. Update valkey_glide.stub.php and valkey_glide_cluster.stub.php with the new configuration options and documentation.

Example PHP usage:

php
$client = new ValkeyGlide();
$client->connect(
    addresses: [['host' => 'localhost', 'port' => 6379]],
    use_tls: true,
    advanced_config: [
        'tls_config' => [
            'root_certs' => $ca_cert_data,
            'client_cert' => $client_cert_pem,
            'client_key' => $client_key_pem,
        ]
    ]
);

Reference implementations:
* [Java GLIDE mTLS](https://github.com/valkey-io/valkey-glide/tree/main/java)
* [Python GLIDE mTLS](https://github.com/valkey-io/valkey-glide/tree/main/python)

### Implementation Details

The glide-core Rust layer already supports mTLS configuration via protobuf fields:
- bytes client_cert = 22; in connection_request.proto
- bytes client_key = 23; in connection_request.proto

The current PHP TLS struct (valkey_glide_tls_advanced_configuration_t in common.h) only has:
c
typedef struct {
    uint8_t* root_certs;
    size_t   root_certs_len;
    bool     use_insecure_tls;
} valkey_glide_tls_advanced_configuration_t;

This needs to be extended with:
c
typedef struct {
    uint8_t* root_certs;
    size_t   root_certs_len;
    uint8_t* client_cert;
    size_t   client_cert_len;
    uint8_t* client_key;
    size_t   client_key_len;
    bool     use_insecure_tls;
} valkey_glide_tls_advanced_configuration_t;

And create_connection_request() in valkey_glide_core_commands.c needs to serialize the new fields into the protobuf message.

### Related
* Other GLIDE clients already support mTLS via glide-core
* The Rust core (glide-core) already has mTLS plumbing available
* C# tracking issue: https://github.com/valkey-io/valkey-glide-csharp/issues/488

### Additional Information
* The glide-core Rust layer already supports mTLS configuration, so the primary work is exposing this capability through the C/Zend extension and the PHP public API.
* This should be a non-breaking addition (new optional configuration keys in the tls_config array).
* Integration tests will require a Valkey server configured with mTLS (client certificate verification enabled).
* The existing stream_context_create(['ssl' => ...]) path in the standalone client could also be extended to pass client cert/key, but the advanced_config path is preferred for consistency across standalone and cluster clients.

prateek-kumar-improving:
### Summary

Adds mutual TLS (mTLS) client certificate support to the Valkey GLIDE PHP client for both standalone (`ValkeyGlide`) and cluster (`ValkeyGlideCluster`) clients, at feature parity with the Java, Python, Node.js, and C# clients. Non-breaking addition — new optional keys under `advanced_config['tls_config']`.

### Issue link

This Pull Request is linked to issue: [feat(config): add mTLS (mutual TLS) support](https://github.com/valkey-io/valkey-glide-php/issues/288)
Closes #288

Two mutually-exclusive mTLS modes under `tls_config` (matching the other GLIDE clients):

- **Byte-based** (static): `client_cert` + `client_key` — inline PEM bytes.
- **Path-based** (with automatic reload): `client_cert_path` + `client_key_path` — the core reads the files and periodically re-reads them, so a rotated certificate is adopted on the next reconnect. Optional `cert_reload_interval_seconds` overrides the reload cadence.

Validation (mirroring the Python reference): each mode requires both of its fields; the two modes are mutually exclusive; byte values must be non-empty and within the max size; paths must be non-empty; `cert_reload_interval_seconds` is path-mode-only, positive, and within uint32; mTLS credentials are rejected when TLS is disabled.

Example usage:

```php
// Byte-based mTLS
$client->connect(
    addresses: [['host' => 'localhost', 'port' => 6379]],
    use_tls: true,
    advanced_config: ['tls_config' => [
        'root_certs'  => file_get_contents('ca-cert.pem'),
        'client_cert' => file_get_contents('client-cert.pem'),
        'client_key'  => file_get_contents('client-key.pem'),
    ]]
);

// Path-based mTLS with automatic reload
$client->connect(
    addresses: [['host' => 'localhost', 'port' => 6379]],
    use_tls: true,
    advanced_config: ['tls_config' => [
        'root_certs'                   => file_get_contents('ca-cert.pem'),
        'client_cert_path'             => 'client-cert.pem',
        'client_key_path'              => 'client-key.pem',
        'cert_reload_interval_seconds' => 300, // optional
    ]]
);
```

### Implementation

- `common.h`: Extended `valkey_glide_tls_advanced_configuration_t` with byte fields (`client_cert`/`client_key`), path fields (`client_cert_path`/`client_key_path`), and `cert_reload_interval` (-1 = unset). Added the config-key constants and `VALKEY_GLIDE_CERTIFICATE_MAX_SIZE` (10 MiB, matching C#'s `CertificateMaxSize`, applied to byte-mode values).
- `valkey_glide.c`: `_parse_mtls_config` implements the two-mode validation from the Python reference (`_validate_mtls_config`): both-or-neither pairing per mode, mode exclusivity, empty/oversized byte rejection, empty-path rejection, and interval range checks. Byte material is copied into buffers; path strings are stored unread (the core reads and reloads them). mTLS with TLS disabled is rejected alongside the existing insecure-TLS guard. Buffers/strings are freed in cleanup.
- `valkey_glide_core_commands.c`: Serializes byte-mode into `client_cert`/`client_key`, and path-mode into `client_cert_path`/`client_key_path` plus a `cert_reload` message (`enabled = true`, optional `interval_seconds`) — matching the Go/Python serialization.
- Updated `valkey_glide.stub.php` and `valkey_glide_cluster.stub.php` docblocks (no signature changes).

Depends on the `valkey-glide` submodule bump (#325), which provides the native `client_cert_path` / `client_key_path` / `cert_reload` protobuf fields.

### Limitations

- mTLS is only configurable via `advanced_config['tls_config']`. Client certificates/keys supplied through a stream context (`context: stream_context_create([...])`) are not used for mTLS. (PHP-specific, since the stream-context TLS path does not exist in the other clients.)

### Behavior notes (consistent with the other GLIDE clients)

- Byte-based mTLS is static (no reload). Automatic certificate rotation is only available with path-based mTLS, where the GLIDE core reads and reloads the files.
- Path-based credentials are validated for presence/pairing at config time, but the files themselves are read by the GLIDE core at connection time — a missing/unreadable/empty path surfaces as a connection error, not a constructor exception.

### Testing

- Unit tests (`tests/ConnectionRequestTest.php`): assert the serialized request for byte-mode (`client_cert`/`client_key`, no reload) and path-mode (`client_cert_path`/`client_key_path` + `cert_reload{enabled, interval}`), plus all validation errors — byte/path both-or-neither, mode exclusivity, empty/oversized byte (cert and key), empty path (cert and key), interval without path mode, interval zero/negative/overflow, and mTLS with TLS disabled. Test matrix mirrors the C#/Python suites. All 81 ConnectionRequest tests pass.
- mTLS handshake integration tests against a client-cert-verifying server, for both standalone (`tests/ValkeyGlideTlsTest.php`, port 6405) and cluster (`tests/ValkeyGlideClusterTlsTest.php`, ports 8101-8106): a successful authenticated connection with byte-based credentials, a successful authenticated connection with path-based credentials, and a rejected connection when no client certificate is presented. The test harness (`tests/start_valkey_with_replicas.sh`, `tests/create-valkey-cluster.sh`) starts the verifying servers via `cluster_manager.py --tls --tls-auth-clients`. This mirrors the Python/other clients' standalone+cluster handshake coverage.
- Verified locally end-to-end: byte- and path-based mTLS produce an authenticated `ping` against a `--tls-auth-clients` server/cluster, and a missing client certificate is rejected.
- `make` builds cleanly with `-Werror`; `./lint-c.sh` (clang-format 18) and `./lint-php.sh` pass.

### Checklist

### DIFF
--- a/tests/ConnectionRequestTest.php
+++ b/tests/ConnectionRequestTest.php
@@ -618,10 +618,291 @@ class ConnectionRequestTest extends \TestSuite
         }
     }
 
+    // mTLS (client cert/key) Tests
+    // -----------------------------
+
+    private const CLIENT_CERT_DATA = "CLIENT_CERT_DATA";
+    private const CLIENT_KEY_DATA  = "CLIENT_KEY_DATA";
+
+    public function testClientAuthTlsDefault()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(use_tls: true);
+        $this->assertEquals('', $request->getClientCert());
+        $this->assertEquals('', $request->getClientKey());
+
+        $request = ClientConstructorMock::simulate_cluster_constructor(use_tls: true);
+        $this->assertEquals('', $request->getClientCert());
+        $this->assertEquals('', $request->getClientKey());
+    }
+
+    public function testClientAuthTlsAdvancedConfig()
+    {
+        $advanced_config = ['tls_config' => [
+            'client_cert' => self::CLIENT_CERT_DATA,
+            'client_key'  => self::CLIENT_KEY_DATA,
+        ]];
+
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+
+        $request = ClientConstructorMock::simulate_cluster_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+    }
+
+    public function testClientAuthTlsWithRootCerts()
+    {
+        $advanced_config = ['tls_config' => [
+            'root_certs'  => self::CERTIFICATE_DATA,
+            'client_cert' => self::CLIENT_CERT_DATA,
+            'client_key'  => self::CLIENT_KEY_DATA,
+        ]];
+
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+
+        $root_certs = $request->getRootCerts();
+        $this->assertEquals(1, $root_certs->count());
+        $this->assertEquals(self::CERTIFICATE_DATA, $root_certs[0]);
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+
+        $request = ClientConstructorMock::simulate_cluster_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+
+        $root_certs = $request->getRootCerts();
+        $this->assertEquals(1, $root_certs->count());
+        $this->assertEquals(self::CERTIFICATE_DATA, $root_certs[0]);
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+    }
+
+    public function testClientAuthTlsCertWithoutKey()
+    {
+        $advanced_config = ['tls_config' => ['client_cert' => self::CLIENT_CERT_DATA]];
+        $expected_msg = 'client_cert is provided but client_key not provided. mTLS requires both.';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertEquals($expected_msg, $e->getMessage());
+        }
+
+        try {
+            ClientConstructorMock::simulate_cluster_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertEquals($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsKeyWithoutCert()
+    {
+        $advanced_config = ['tls_config' => ['client_key' => self::CLIENT_KEY_DATA]];
+        $expected_msg = 'client_key is provided but client_cert not provided. mTLS requires both.';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertEquals($expected_msg, $e->getMessage());
+        }
+
+        try {
+            ClientConstructorMock::simulate_cluster_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertEquals($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsEmptyCert()
+    {
+        $advanced_config = ['tls_config' => [
+            'client_cert' => '',
+            'client_key'  => self::CLIENT_KEY_DATA,
+        ]];
+        $expected_msg = 'client_cert cannot be an empty string';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsEmptyKey()
+    {
+        $advanced_config = ['tls_config' => [
+            'client_cert' => self::CLIENT_CERT_DATA,
+            'client_key'  => '',
+        ]];
+        $expected_msg = 'client_key cannot be an empty string';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsOversizedCert()
+    {
+        // 10 MiB + 1 byte exceeds the maximum allowed certificate size.
+        $oversized = str_repeat('A', (10 * 1024 * 1024) + 1);
+        $advanced_config = ['tls_config' => [
+            'client_cert' => $oversized,
+            'client_key'  => self::CLIENT_KEY_DATA,
+        ]];
+        $expected_msg = 'client_cert exceeds the maximum allowed size';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsOversizedKey()
+    {
+        $oversized = str_repeat('A', (10 * 1024 * 1024) + 1);
+        $advanced_config = ['tls_config' => [
+            'client_cert' => self::CLIENT_CERT_DATA,
+            'client_key'  => $oversized,
+        ]];
+        $expected_msg = 'client_key exceeds the maximum allowed size';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsFromFilePath()
+    {
+        // Write cert and key to temporary files and load them via *_path options.
+        $cert_handle = tmpfile();
+        fwrite($cert_handle, self::CLIENT_CERT_DATA);
+        $cert_path = stream_get_meta_data($cert_handle)['uri'];
+
+        $key_handle = tmpfile();
+        fwrite($key_handle, self::CLIENT_KEY_DATA);
+        $key_path = stream_get_meta_data($key_handle)['uri'];
+
+        $advanced_config = ['tls_config' => [
+            'client_cert_path' => $cert_path,
+            'client_key_path'  => $key_path,
+        ]];
+
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+
+        $request = ClientConstructorMock::simulate_cluster_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+
+        fclose($cert_handle);
+        fclose($key_handle);
+    }
+
+    public function testClientAuthTlsMixedInlineAndPath()
+    {
+        $key_handle = tmpfile();
+        fwrite($key_handle, self::CLIENT_KEY_DATA);
+        $key_path = stream_get_meta_data($key_handle)['uri'];
+
+        // Inline cert bytes with key loaded from file path.
+        $advanced_config = ['tls_config' => [
+            'client_cert'     => self::CLIENT_CERT_DATA,
+            'client_key_path' => $key_path,
+        ]];
+
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            use_tls: true,
+            advanced_config: $advanced_config
+        );
+        $this->assertEquals(self::CLIENT_CERT_DATA, $request->getClientCert());
+        $this->assertEquals(self::CLIENT_KEY_DATA, $request->getClientKey());
+
+        fclose($key_handle);
+    }
+
+    public function testClientAuthTlsPathNotFound()
+    {
+        $advanced_config = ['tls_config' => [
+            'client_cert_path' => '/invalid/client-cert.pem',
+            'client_key'       => self::CLIENT_KEY_DATA,
+        ]];
+        $expected_msg = 'Failed to load client_cert from file';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsEmptyPath()
+    {
+        $advanced_config = ['tls_config' => [
+            'client_cert_path' => '',
+            'client_key'       => self::CLIENT_KEY_DATA,
+        ]];
+        $expected_msg = 'client_cert_path cannot be an empty string';
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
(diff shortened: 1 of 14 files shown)


valkey_glide.c
```diff
@@ -1818,6 +1839,147 @@ static bool _determine_use_tls(valkey_glide_php_common_constructor_params_t* par
...
+            snprintf(
+                error_message, sizeof(error_message), "%s cannot be an empty string.", path_key);
+            VALKEY_LOG_ERROR("tls_config_mtls", error_message);
+            zend_throw_exception(get_valkey_glide_exception_ce(), error_message, 0);
+            return false;
+        }
+
+        uint8_t* file_data;
+        size_t   file_len;
+        if (!_load_data_from_file(path, &file_data, &file_len)) {
```
xShinnRyuu: Minor (security/stability): the 10 MiB size cap is enforced only after the whole file is read. `_load_data_from_file` does `emalloc(st_size)` plus a full `fread` (around line 2201-2205), and the `file_len > VALKEY_GLIDE_CERTIFICATE_MAX_SIZE` check happens afterward at line 1960. So an accidental multi-GB or wrong file (exactly what the constant's comment says it guards against) is fully loaded into memory before rejection. The inline path checks the cap before copying, so only the file path has this gap. Consider checking `st.st_size` against the cap inside `_load_data_from_file` before allocating/reading (for example via an optional max-size parameter), which would also improve the existing `root_certs` cafile path. Low severity because the path is developer-supplied config, not attacker input.

valkey_glide.c
```diff
@@ -1895,6 +2061,50 @@ static valkey_glide_tls_advanced_configuration_t* _build_advanced_tls_config(
             tls_advanced_config->root_certs     = emalloc(cert_len);
             memcpy(tls_advanced_config->root_certs, Z_STRVAL_P(root_certs), cert_len);
         }
+
+        // Client certificate (mTLS) - inline PEM bytes or file path.
```
xShinnRyuu: Nit (completeness): mTLS material is resolved only inside the `advanced_tls_ht` branch. When TLS is configured through a PHP stream context instead (the other supported TLS path in this function), `client_cert`/`client_key`/`client_cert_path`/`client_key_path` are never read and are silently ignored. Consider either documenting that mTLS requires `advanced_config['tls_config']`, or rejecting these keys when a stream context is in use so a misconfiguration is not silently dropped.

tests/ConnectionRequestTest.php
```diff
@@ -618,10 +618,315 @@ public function testRootCertsAdvancedConfigAndStreamContext()
...
+
+        try {
+            ClientConstructorMock::simulate_standalone_constructor(use_tls: true, advanced_config: $advanced_config);
+            $this->assertTrue(false, 'Expected ValkeyGlideException was not thrown');
+        } catch (ValkeyGlideException $e) {
+            $this->assertStringContains($expected_msg, $e->getMessage());
+        }
+    }
+
+    public function testClientAuthTlsOversizedCert()
```
xShinnRyuu: Nit (test coverage): the oversize tests cover inline cert/key only. The file-path oversize branch (`file_len > MAX`, the `efree(file_data)`+throw at valkey_glide.c:1960-1972) is not exercised, and there is no assertion that exactly 10 MiB is accepted (the cap is `> MAX`, so the inclusive boundary is untested). The symmetric key-path negatives (empty `client_key_path`, `client_key`+`client_key_path` conflict) are also untested, though they share the same `_resolve_mtls_material` code as the cert cases.

valkey_glide_core_commands.c
```diff
@@ -121,6 +121,16 @@ uint8_t* create_connection_request(size_t*                                   len
             conn_req.n_root_certs = 1;
             conn_req.root_certs   = &root_cert_data;
         }
+
+        /* Set client certificate and key for mutual TLS (mTLS) */
```
xShinnRyuu: **mTLS parity note (cross-repo review against `valkey-io/valkey-glide`): path-based mTLS diverges from the glide-core contract.**

This block only ever populates the inline byte fields `client_cert` (proto field 22) and `client_key` (field 23). That part is correct and at full parity with the byte-field contract and the Python reference binding, whose `client_cert_pem` / `client_key_pem` also land on 22/23.

The divergence is with the `client_cert_path` / `client_key_path` config keys. In glide-core and the reference bindings these are native proto fields (`client_cert_path = 31`, `client_key_path = 32`, `cert_reload = 33`): the client passes the path unread (Python uses `os.fspath`), and the CORE reads the files at connection time and, with reload enabled, periodically re-reads them so a rotated cert is adopted on the next reconnect. Here PHP instead reads the file client-side via `_load_data_from_file` and lands the static bytes on fields 22/23, so PHP's `*_path` is a one-time read with no core-side reload. Same key name, materially different semantics, and the stub docs do not call this out.

Root cause is version skew, not a coding mistake: the pinned `valkey-glide` submodule (`1b0ffb14`, 2026-06-26) predates fields 31/32/33, which were added upstream on 2026-07-21 in #6386. This generated proto simply has no path/reload fields to set, so client-side reading is the only option at this pin.

Suggestion: inline byte mTLS is fine to ship as-is. For the `*_path` keys, consider either (a) documenting explicitly that they are a client-side static read with no reload, distinct from the core's native path fields; or (b) deferring them until the submodule is bumped to a core with fields 31/32/33 and wiring native paths plus `cert_reload` for true parity. Option (b) avoids a later breaking change when the same-named keys switch from client-side bytes to native path semantics.

valkey_glide.c
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/12f6c7dddb09ac4d0dea54030fc8fbde66e2a1dd

`_load_data_from_file` now takes a `max_size` parameter and rejects files larger than `VALKEY_GLIDE_CERTIFICATE_MAX_SIZE` (10 MiB) right after `fstat`, before any `emalloc`/`fread`. The bounded loader is used for `client_cert_path`, `client_key_path`, and the `root_certs` cafile path. Added `testClientAuthTlsOversizedCertFromPath` to cover the file-path oversize branch.

valkey_glide.c
prateek-kumar-improving: Documented in this commit: https://github.com/valkey-io/valkey-glide-php/commit/73d9e579ced7dbb0884cd6f9ba6499d1e1edb5bb

mTLS is now explicitly documented as only supported via `advanced_config['tls_config']`. Added notes to the `connect()` stubs (standalone and cluster) on both the `$advanced_config` tls_config block and the `$context` parameter, plus a note in the README mTLS section stating that client certificates supplied through a stream context are not used for mutual TLS.

tests/ConnectionRequestTest.php
prateek-kumar-improving: Addressed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/2080c852a3ccf0a584c8ee1611a10dbf6b0c5e0b

Added:
- `testClientAuthTlsBoundarySizeAccepted`
- `testClientAuthTlsOversizedKeyFromPath`
- `testClientAuthTlsEmptyKeyPath`
- `testClientAuthTlsKeyInlineAndPathConflict`

(cert-side file-path oversize was already covered by `testClientAuthTlsOversizedCertFromPath`.)

valkey_glide_core_commands.c
prateek-kumar-improving: Follow-up issue created: https://github.com/valkey-io/valkey-glide-php/issues/324

Confirmed the divergence you noted: the pinned submodule (`1b0ffb14`) predates the native path fields (`client_cert_path = 31`, `client_key_path = 32`, `cert_reload = 33`), so this PR keeps the inline byte fields (22/23) at full parity and treats `*_path` as a client-side read. Bumping the submodule to pick up the native fields is blocked on updating `utils/patch_proto_and_rust.py` for newer `optional` fields; #324 tracks that bump plus the follow-up to convert `*_path` to native core-side read + `cert_reload` semantics (mirroring the Python reference).

valkey_glide.c
```diff
@@ -1895,6 +2052,50 @@ static valkey_glide_tls_advanced_configuration_t* _build_advanced_tls_config(
...
+        if (!_resolve_mtls_material(advanced_tls_ht,
+                                    VALKEY_GLIDE_CLIENT_CERT,
```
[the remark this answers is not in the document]
prateek-kumar-improving: Already updated.

valkey_glide.c
xShinnRyuu: Why did we create another parameter field to pass in a constant that can be used in the function to check the max size directly?

valkey_glide.c
```diff
@@ -1827,6 +1979,10 @@ static valkey_glide_tls_advanced_configuration_t* _build_advanced_tls_config(
     tls_advanced_config->use_insecure_tls = false;
     tls_advanced_config->root_certs       = NULL;
     tls_advanced_config->root_certs_len   = 0;
+    tls_advanced_config->client_cert      = NULL;
+    tls_advanced_config->client_cert_len  = 0;
+    tls_advanced_config->client_key       = NULL;
+    tls_advanced_config->client_key_len   = 0;
```
xShinnRyuu: ecalloc already zero-fills the allocation, so these four assignments are dead and can be removed.

valkey_glide.c
```diff
@@ -1818,6 +1839,137 @@ static bool _determine_use_tls(valkey_glide_php_common_constructor_params_t* par
...
+        }
+        *out_len  = len;
+        *out_data = emalloc(len);
+        memcpy(*out_data, Z_STRVAL_P(inline_val), len);
+        return true;
+    }
+
+    if (has_path) {
+        const char* path = Z_STRVAL_P(path_val);
+        if (Z_STRLEN_P(path_val) == 0) {
```
xShinnRyuu: This check is redundant. An empty path already fails in _load_data_from_file because fopen("") returns NULL, so this branch only changes the error message. You can drop it and leave the empty-path case to the loader.

valkey_glide.c
prateek-kumar-improving: Since, it is general static helper method, thought of keeping it as a parameter, in case it would have other uses.

xShinnRyuu (approved):

valkey_glide.c
xShinnRyuu: I tend to not try to future proof internal methods like these. Anyone else that looks at this code and sees this being passed through seems odd, but this is more of a design decision so feel free to resolve this if you dont want to change it

valkey_glide.c
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-php/commit/6de20a25e374424f7ba8e80d8fb9e65dcd420db4 — removed the redundant zero-initialization assignments since `ecalloc` already zero-fills the allocation.

valkey_glide.c
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-php/commit/6de20a25e374424f7ba8e80d8fb9e65dcd420db4 — dropped the explicit empty-path branch and let `_load_data_from_file` handle it (`fopen("")` fails, so the loader reports the error). Updated `testClientAuthTlsEmptyPath` / `testClientAuthTlsEmptyKeyPath` to expect the loader's message.

valkey_glide.c
prateek-kumar-improving: Agreed — removed the `max_size` parameter in https://github.com/valkey-io/valkey-glide-php/commit/6de20a25e374424f7ba8e80d8fb9e65dcd420db4. `_load_data_from_file` now references `VALKEY_GLIDE_CERTIFICATE_MAX_SIZE` directly, and both call sites (mTLS path and the `cafile` root cert) are simplified.

valkey_glide.c
```diff
@@ -1818,15 +1849,162 @@ static bool _determine_use_tls(valkey_glide_php_common_constructor_params_t* par
...
+        advanced_tls_ht, VALKEY_GLIDE_CLIENT_KEY_PATH, sizeof(VALKEY_GLIDE_CLIENT_KEY_PATH) - 1);
+    zval* interval_zv = zend_hash_str_find(advanced_tls_ht,
+                                           VALKEY_GLIDE_CERT_RELOAD_INTERVAL,
+                                           sizeof(VALKEY_GLIDE_CERT_RELOAD_INTERVAL) - 1);
+
+    bool has_cert_pem  = cert_pem && Z_TYPE_P(cert_pem) == IS_STRING;
+    bool has_key_pem   = key_pem && Z_TYPE_P(key_pem) == IS_STRING;
+    bool has_cert_path = cert_path && Z_TYPE_P(cert_path) == IS_STRING;
+    bool has_key_path  = key_path && Z_TYPE_P(key_path) == IS_STRING;
+    bool has_interval  = interval_zv != NULL;
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/cb5d365aa28ec5179c628ee2ba5a64932ddfe20a — `_parse_mtls_config` now rejects non-string credential fields and a non-integer `cert_reload_interval_seconds` up front (instead of treating wrong types as absent or coercing). Added `testClientAuthTlsWrongTypedCredential` and `testClientAuthTlsWrongTypedInterval`.

tests/ValkeyGlideTlsTest.php
```diff
@@ -265,4 +265,183 @@ public function testTlsWithIPv6AddressStream()
...
+            $client = new ValkeyGlide();
+            $client->connect(
+                addresses: [self::TLS_ADDRESS_STANDALONE],
+                use_tls: true,
+                advanced_config: [
+                    'tls_config' => [
+                        'client_cert'      => $certData,
+                        'client_key'       => $certData,
+                        'client_cert_path' => self::TLS_CERTIFICATE_PATH,
+                        'client_key_path'  => self::TLS_CERTIFICATE_PATH,
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/cb5d365aa28ec5179c628ee2ba5a64932ddfe20a — both mutual-exclusion tests (standalone and cluster) now load real client certificate/key material from `TLS_CLIENT_CERT_PATH`/`TLS_CLIENT_KEY_PATH` for the inline and path values, and set `root_certs`.

tests/start_valkey_with_replicas.sh
```diff
@@ -72,6 +72,14 @@ else
     echo "⚠️  WARNING: TLS standalone setup failed (port 6400 may be in use), continuing without TLS..."
 fi
 
+# Handle mTLS setup (client-certificate verification) with graceful failure
+echo "Setting up mTLS standalone server (client-cert verification)..."
+if ../valkey-glide/utils/cluster_manager.py --tls start --tls-auth-clients --prefix mtls-standalone -p 6405 -r 0; then
+    echo "✅ mTLS standalone server started on port 6405"
+else
+    echo "⚠️  WARNING: mTLS standalone setup failed (port 6405 may be in use), continuing without mTLS..."
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/cb5d365aa28ec5179c628ee2ba5a64932ddfe20a — kept the harness's graceful-continue convention but added a `skipIfMtlsUnavailable($port)` guard so the mTLS handshake tests skip (rather than fail) when the client-cert-verifying server did not start. Verified: server up -> tests run/pass; server down -> tests skip.

valkey_glide.c
```diff
@@ -1895,6 +2036,50 @@ static valkey_glide_tls_advanced_configuration_t* _build_advanced_tls_config(
             tls_advanced_config->root_certs     = emalloc(cert_len);
             memcpy(tls_advanced_config->root_certs, Z_STRVAL_P(root_certs), cert_len);
         }
+
+        // Client certificate (mTLS) - inline PEM bytes or file path.
+        if (!_resolve_mtls_material(advanced_tls_ht,
+                                    VALKEY_GLIDE_CLIENT_CERT,
```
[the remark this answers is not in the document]
prateek-kumar-improving: This is already handled: `valkey_glide_build_client_config_base` rejects mTLS credentials (byte and path fields) when TLS is disabled, throwing "Cannot configure mTLS client certificate when TLS is disabled." before the request is serialized — so it never proceeds with `NoTls`. Covered by `testClientAuthTlsWithTlsDisabled`.

yipin-chen (approved):

PR #322, opened 2026-08-25
feat(standalone): add NodeDiscoveryMode support for standalone clients

### LINKED ISSUE #302: [Task] Add NodeDiscoveryMode support for standalone clients
## Description

Expose the `NodeDiscoveryMode` configuration option for standalone clients, matching the implementation already available in Java, Python, Node, and Go.

## Modes

- **STANDARD** (default): Current behavior — verifies node roles via `INFO REPLICATION`, uses only provided addresses.
- **STATIC**: Skips role detection. Trusts provided addresses as-is (first = primary). Enables connectivity through proxies (Envoy, HAProxy) that don't support `INFO`.
- **DISCOVER_ALL**: Auto-discovers full topology from any single starting node (primary or replica).

## Tasks

- [ ] Expose `NodeDiscoveryMode` enum/constants in PHP
- [ ] Add `node_discovery_mode` parameter to `ValkeyGlide::connect()`
- [ ] Wire config to protobuf field 27 on `ConnectionRequest`
- [ ] Add integration tests for STATIC and DISCOVER_ALL modes
- [ ] Update documentation

## Reference

- valkey-io/valkey-glide#5724 — Core implementation (merged)
- valkey-io/valkey-glide#5091 — Original feature request
- valkey-io/valkey-glide#4670 — Related issue

prateek-kumar-improving:
### Summary

Exposes the `NodeDiscoveryMode` configuration option for standalone clients, bringing the PHP client in line with Java, Python, Node, Go, and C#. This controls how the client discovers node roles and topology in standalone mode.

### Issue link

This Pull Request is linked to issue:
Closes #302

Adds a `node_discovery_mode` parameter to `ValkeyGlide::connect()` and three new class constants:

- `ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD` (default) — Verifies node roles via `INFO REPLICATION` and uses only the provided addresses. Existing behaviour; no change for current users.
- `ValkeyGlide::NODE_DISCOVERY_MODE_STATIC` — Skips role detection and trusts the provided addresses as-is (first address treated as primary). Enables connectivity through proxies such as Envoy/HAProxy that do not support `INFO`.
- `ValkeyGlide::NODE_DISCOVERY_MODE_DISCOVER_ALL` — Auto-discovers the full topology (primary + all replicas) from any single starting node (primary or replica).

The parameter is optional and defaults to `STANDARD`, so this change is fully backward compatible.

### Implementation

- Added `valkey_glide_node_discovery_mode_t` enum and a `node_discovery_mode` field to the base client configuration and constructor-params structs (`common.h`).
- Parsed the new `node_discovery_mode` argument in `ValkeyGlide::connect()` and mapped it into the client configuration (`valkey_glide.c`).
- Wired the configuration to the protobuf `node_discovery_mode` field on `ConnectionRequest` in `create_connection_request()`, standalone-only (`valkey_glide_core_commands.c`).
- Exposed the constants and the `connect()` parameter in the PHP stub (`valkey_glide.stub.php`); arginfo is regenerated by the build.
- Extended `ClientConstructorMock::simulate_standalone_constructor()` to support the option for serialization tests.

Notes:
- The protobuf field is `node_discovery_mode = 28` on `ConnectionRequest` (field 27 is `client_side_cache`), not field 27 as stated in the issue.
- The generated protobuf PHP class `tests/Connection_request/NodeDiscoveryMode.php` already existed in the repo (the proto had been regenerated previously), so it does not appear in this diff.

### Limitations

- `NodeDiscoveryMode` applies to standalone clients only; it is intentionally not set for cluster clients. This matches the reference clients (confirmed in glide-core, where only the standalone client path reads the field) and the C# client, whose docs state the option is "ignored in cluster mode".
- The PHP client does not expose a `read_only` connection flag, so the `read_only` + `DISCOVER_ALL` rejection test present in the Java/Python/Node suites is not applicable here and was not ported. Adding `read_only` support is tracked separately in #323.

### Testing

- 4 protobuf serialization tests in `ConnectionRequestTest` verifying each mode maps to the correct `ConnectionRequest.node_discovery_mode` value (plus the default). These mirror the C# unit tests in `ConnectionConfigurationTests.cs`.
- 5 integration tests in a new `ValkeyGlideNodeDiscoveryModeTest` covering STATIC (connect/read, writes) and DISCOVER_ALL (from primary, from replica, from partial addresses), using the primary/replica topology from `tests/start_valkey_with_replicas.sh`. Coverage matches the Java/Python/Node discovery suites. The tests self-skip when the topology or TLS-only runs make them inapplicable.
- Cross-checked the implementation against the C# client (`valkey-io/valkey-glide-csharp`): enum discriminants (`Standard=0`, `Static=1`, `DiscoverAll=2`), standalone-only scoping, and default behaviour all match.
- Verified locally: build passes with `-Werror`; C lint passes (clang-format 18); PHP lint passes; ConnectionRequest suite 64/64; NodeDiscoveryMode integration 5/5; existing constructor/connect tests 21/21.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,30 @@ $client->connect(
 )
 ```
 
+### Node Discovery Mode (Standalone)
+
+For standalone clients, `node_discovery_mode` controls how the client discovers node roles and topology. It is only applicable to standalone clients (`ValkeyGlide`).
+
+- `ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD` (default): Verifies node roles via `INFO REPLICATION` and uses only the provided addresses.
+- `ValkeyGlide::NODE_DISCOVERY_MODE_STATIC`: Skips role detection. Trusts the provided addresses as-is (the first address is treated as the primary). Enables connectivity through proxies (e.g., Envoy, HAProxy) that do not support `INFO`. Do not set `client_name` when using this mode with a proxy.
+- `ValkeyGlide::NODE_DISCOVERY_MODE_DISCOVER_ALL`: Auto-discovers the full topology (primary + all replicas) from any single starting node (primary or replica).
+
+```php
+// STATIC: trust provided addresses, skip role detection (e.g., behind a proxy)
+$client = new ValkeyGlide();
+$client->connect(
+    addresses: [['host' => 'proxy-host', 'port' => 6379]],
+    node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_STATIC
+);
+
+// DISCOVER_ALL: discover the full topology from any single node
+$client = new ValkeyGlide();
+$client->connect(
+    addresses: [['host' => 'localhost', 'port' => 6379]],
+    node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_DISCOVER_ALL
+);
+```
+
 ### Cluster Valkey
 
 ```php
--- a/tests/ConnectionRequestTest.php
+++ b/tests/ConnectionRequestTest.php
@@ -81,6 +81,7 @@ require_once __DIR__ . "/Connection_request/CompressionConfig.php";
 require_once __DIR__ . "/Connection_request/ConnectionRequest.php";
 require_once __DIR__ . "/Connection_request/ConnectionRetryStrategy.php";
 require_once __DIR__ . "/Connection_request/NodeAddress.php";
+require_once __DIR__ . "/Connection_request/NodeDiscoveryMode.php";
 require_once __DIR__ . "/Connection_request/PeriodicChecksDisabled.php";
 require_once __DIR__ . "/Connection_request/PeriodicChecksManualInterval.php";
 require_once __DIR__ . "/Connection_request/ProtocolVersion.php";
@@ -210,6 +211,36 @@ class ConnectionRequestTest extends \TestSuite
     {
         $request = ClientConstructorMock::simulate_cluster_constructor(read_from: ValkeyGlide::READ_FROM_AZ_AFFINITY_REPLICAS_AND_PRIMARY);
         $this->assertEquals(\Connection_request\ReadFrom::AZAffinityReplicasAndPrimary, $request->getReadFrom());
+    }
+
+    public function testStandaloneNodeDiscoveryModeDefault()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor();
+        $this->assertEquals(\Connection_request\NodeDiscoveryMode::Standard, $request->getNodeDiscoveryMode());
+    }
+
+    public function testStandaloneNodeDiscoveryModeStandard()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD
+        );
+        $this->assertEquals(\Connection_request\NodeDiscoveryMode::Standard, $request->getNodeDiscoveryMode());
+    }
+
+    public function testStandaloneNodeDiscoveryModeStatic()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_STATIC
+        );
+        $this->assertEquals(\Connection_request\NodeDiscoveryMode::PBStatic, $request->getNodeDiscoveryMode());
+    }
+
+    public function testStandaloneNodeDiscoveryModeDiscoverAll()
+    {
+        $request = ClientConstructorMock::simulate_standalone_constructor(
+            node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_DISCOVER_ALL
+        );
+        $this->assertEquals(\Connection_request\NodeDiscoveryMode::DiscoverAll, $request->getNodeDiscoveryMode());
     }
 
     public function testStandaloneRequestTimeout()
--- a/tests/ConnectionRequestTest.php
+++ b/tests/ConnectionRequestTest.php
@@ -243,6 +243,13 @@ class ConnectionRequestTest extends \TestSuite
         $this->assertEquals(\Connection_request\NodeDiscoveryMode::DiscoverAll, $request->getNodeDiscoveryMode());
     }
 
+    public function testStandaloneNodeDiscoveryModeInvalidThrows()
+    {
+        $this->assertThrows(ValkeyGlideException::class, function () {
+            ClientConstructorMock::simulate_standalone_constructor(node_discovery_mode: 99);
+        });
+    }
+
     public function testStandaloneRequestTimeout()
     {
         $request = ClientConstructorMock::simulate_standalone_constructor(request_timeout: 999);
--- a/tests/TestValkeyGlide.php
+++ b/tests/TestValkeyGlide.php
@@ -101,6 +101,7 @@ require_once __DIR__ . "/ClientSideCacheTest.php";
 require_once __DIR__ . "/AddressResolverTest.php";
 require_once __DIR__ . "/AddressResolverClusterTest.php";
 require_once __DIR__ . "/ValkeyGlideMonitorTest.php";
+require_once __DIR__ . "/ValkeyGlideNodeDiscoveryModeTest.php";
 
 function getClassArray($classes)
 {
@@ -141,6 +142,7 @@ function getTestClass($class)
         'addressresolver' => 'AddressResolverTest',
         'addressresolvercluster' => 'AddressResolverClusterTest',
         'valkeyglidemonitor' => 'ValkeyGlideMonitorTest',
+        'valkeyglidenodediscoverymode' => 'ValkeyGlideNodeDiscoveryModeTest',
     ];
 
     /* Return early if the class is one of our built-in ones */
--- a/tests/TestValkeyGlide.php
+++ b/tests/TestValkeyGlide.php
@@ -192,6 +192,7 @@ $default_classes = implode(',', [
     'addressresolver',
     'addressresolvercluster',
     'valkeyglidemonitor',
+    'valkeyglidenodediscoverymode',
 ]);
 
 $classes = getClassArray($opt['class'] ?? $default_classes);
--- a/tests/ValkeyGlideNodeDiscoveryModeTest.php
+++ b/tests/ValkeyGlideNodeDiscoveryModeTest.php
@@ -76,6 +76,34 @@ class ValkeyGlideNodeDiscoveryModeTest extends ValkeyGlideBaseTest
     }
 
     /* ----------------------------------------------------------------------
+     * STANDARD mode (default)
+     * ------------------------------------------------------------------- */
+
+    public function testStandardModeConnectsAndReads()
+    {
+        // STANDARD is the default mode: it verifies node roles via INFO REPLICATION
+        // and uses only the provided addresses. Connecting to the primary should
+        // succeed and allow reads and writes.
+        $client = new ValkeyGlide();
+        $client->connect(
+            addresses: [['host' => '127.0.0.1', 'port' => self::PRIMARY_PORT]],
+            request_timeout: 2000,
+            node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_STANDARD
+        );
+
+        $this->assertConnected($client);
+
+        $key = 'ndm_standard_' . uniqid();
+        try {
+            $this->assertTrue($client->set($key, 'value'));
+            $this->assertEquals('value', $client->get($key));
+        } finally {
+            $client->del($key);
+            $client->close();
+        }
+    }
+
+    /* ----------------------------------------------------------------------
      * STATIC mode
      * ------------------------------------------------------------------- */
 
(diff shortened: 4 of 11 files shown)


tests/TestValkeyGlide.php
```diff
@@ -141,6 +142,7 @@ function getTestClass($class)
         'addressresolver' => 'AddressResolverTest',
         'addressresolvercluster' => 'AddressResolverClusterTest',
         'valkeyglidemonitor' => 'ValkeyGlideMonitorTest',
+        'valkeyglidenodediscoverymode' => 'ValkeyGlideNodeDiscoveryModeTest',
```
jamesx-improving, line 145: This registers the class in `$valid_classes`, but it's missing from `$default_classes` (lines 178-195) and there's no dedicated `--class` CI step. The main CI job runs `TestValkeyGlide.php` with no `--class` (test.yml, php.yml), so the 5 new STATIC/DISCOVER_ALL integration tests never execute — a coverage regression vs. the Node reference binding, which runs its NodeDiscoveryMode suite. (greptile flagged this too.)

**Suggested fix:** add `'valkeyglidenodediscoverymode'` to `$default_classes`, or add a dedicated `--class=ValkeyGlideNodeDiscoveryModeTest` CI step alongside the Json/Search ones.

jamesx-improving (commented):
This faithfully ports the monorepo NodeDiscoveryMode feature — enum values, protobuf field/encoding, default, and standalone-only scoping all match the Rust core and the Node binding. One open item flagged inline: the new integration test class isn't wired into the default CI runner, so it never executes.

jamesx-improving (approved):

valkey_glide.stub.php
```diff
@@ -417,6 +445,7 @@ public function connect(
         ?array $client_side_cache = null,
         ?callable $address_resolver = null,
         ?array $circuit_breaker = null,
+        ?int $node_discovery_mode = null,
```
currantw, line 471: Why not explictly set the default here?


```suggestion
        int $node_discovery_mode = NODE_DISCOVERY_MODE_STANDARD,
```

valkey_glide_core_commands.c
```diff
@@ -150,6 +150,17 @@ uint8_t* create_connection_request(size_t*                                   len
...
+    /* Map node_discovery_mode configuration (standalone only). */
+    if (!is_cluster) {
+        if (config->node_discovery_mode == VALKEY_GLIDE_NODE_DISCOVERY_MODE_STATIC) {
+            conn_req.node_discovery_mode = CONNECTION_REQUEST__NODE_DISCOVERY_MODE__Static;
+        } else if (config->node_discovery_mode == VALKEY_GLIDE_NODE_DISCOVERY_MODE_DISCOVER_ALL) {
+            conn_req.node_discovery_mode = CONNECTION_REQUEST__NODE_DISCOVERY_MODE__DiscoverAll;
+        } else {
+            conn_req.node_discovery_mode = CONNECTION_REQUEST__NODE_DISCOVERY_MODE__Standard;
+        }
+    }
```
currantw, line 198: In general, I don't think that we should silently fall through to a default value for enums: it's error prone if additional values are ever added to the enum.

```suggestion
    if (!is_cluster) {
        if (config->node_discovery_mode == VALKEY_GLIDE_NODE_DISCOVERY_MODE_STATIC) {
            conn_req.node_discovery_mode = CONNECTION_REQUEST__NODE_DISCOVERY_MODE__Static;
        } else if (config->node_discovery_mode == VALKEY_GLIDE_NODE_DISCOVERY_MODE_DISCOVER_ALL) {
            conn_req.node_discovery_mode = CONNECTION_REQUEST__NODE_DISCOVERY_MODE__DiscoverAll;
        } else if {
            conn_req.node_discovery_mode = CONNECTION_REQUEST__NODE_DISCOVERY_MODE__Standard;
        } else {
            zend_throw_exception(get_valkey_glide_exception_ce(), "Invalid node_discovery_mode.", 0);
            *len = 0;
            return NULL;
        }
    }
```

I think we should also do the same for `read_from` above.

valkey_glide.c
```diff
@@ -1118,34 +1135,35 @@ static int valkey_glide_create_connection(valkey_glide_object* valkey_glide,
...
+    size_t client_name_len          = 0;
+    char*  client_az                = NULL;
+    size_t client_az_len            = 0;
+    zval*  advanced_config          = NULL;
+    zval*  lazy_connect_zval        = NULL;
+    zval*  context                  = NULL;
+    zval*  compression              = NULL;
+    zval*  client_side_cache        = NULL;
+    zval*  address_resolver         = NULL;
+    zval*  circuit_breaker          = NULL;
```
currantw, line 1204: Why is the order of parameters here different than elsewhere? As much as possible, I think we should try to keep these consistent? Otherwise, especially as the number of parameters grows, it can be difficult to actually find what you're looking for, and makes errors more likely.

valkey_glide.c
```diff
@@ -233,6 +234,20 @@ int valkey_glide_build_client_config_base(valkey_glide_php_common_constructor_pa
...
+            config->node_discovery_mode = VALKEY_GLIDE_NODE_DISCOVERY_MODE_STATIC;
+            break;
+        case 2: /* DISCOVER_ALL */
+            config->node_discovery_mode = VALKEY_GLIDE_NODE_DISCOVERY_MODE_DISCOVER_ALL;
+            break;
+        case 0: /* STANDARD */
+        default:
+            config->node_discovery_mode = VALKEY_GLIDE_NODE_DISCOVERY_MODE_STANDARD;
+            break;
+    }
```
currantw, line 259: Similar to my other comment, I do not think that we should silently fall through to standard mode. At the very least, we should raise an error before defaulting. Same for `read_from` above.

tests/ConnectionRequestTest.php
currantw, line 1: To follow up on other comments, I think we should also test the case where an invalid mode is provided?

tests/ValkeyGlideNodeDiscoveryModeTest.php
```diff
@@ -0,0 +1,212 @@
...
+    {
+        $client = new ValkeyGlide();
+        $client->connect(
+            addresses: [['host' => '127.0.0.1', 'port' => self::PRIMARY_PORT]],
+            request_timeout: 2000,
+            node_discovery_mode: ValkeyGlide::NODE_DISCOVERY_MODE_STATIC
+        );
+
+        // A missing key returns the client's nil value (false for this client).
+        $this->assertEquals($this->getNilValue(), $client->get('nonexistent_' . uniqid()));
```
currantw: Why did you choose this to test? Do we have an existing `assertConnected` we could use instead? Or are you specifically trying to test a missing key? 🤔

tests/ValkeyGlideNodeDiscoveryModeTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+        $client->close();
+    }
+
+    public function testStaticModeAllowsWrites()
```
currantw, line 160: Since the setup is identical, might be easier to just combine this with the test case above? 🤷

tests/ValkeyGlideNodeDiscoveryModeTest.php
```diff
@@ -0,0 +1,212 @@
...
+        try {
+            $this->waitFor(function () use ($probe, $uniqueName) {
+                $clientList = (string) $probe->rawcommand('CLIENT', 'LIST');
+                return str_contains($clientList, $uniqueName);
+            }, 10, "Discovery client '{$uniqueName}' was not found on the replica");
+        } finally {
+            $probe->close();
+            $client->close();
+        }
+    }
```
currantw, line 210: This is a cool way to test this! Nice! (You can obviously resolve)

tests/ValkeyGlideNodeDiscoveryModeTest.php
```diff
@@ -0,0 +1,212 @@
...
+        );
+
+        $key = 'ndm_discover_replica_key_' . uniqid();
+
+        $probe = $this->connectProbe(self::REPLICA0_PORT);
+
+        try {
+            // Writes must be routed to the discovered primary.
+            $this->assertTrue($client->set($key, 'value'));
+            $this->assertEquals('value', $client->get($key));
```
currantw, line 233: What do these writes do here? Are they necessary to force the discovery of the primary?

tests/ValkeyGlideNodeDiscoveryModeTest.php
currantw, line 1: Probably worth adding a basic STANDARD test for completeness.

README.md
```diff
@@ -379,6 +379,30 @@ $client->connect(
 )
 ```
 
+### Node Discovery Mode (Standalone)
```
currantw, line 433: I don't think we should generally duplicate this documentation here. Rather, it should be centralized on the docs site:

1. Do you have any issue to track this to make sure that gets added?
2. Does the existing code here link to that webpage?

currantw (commented):
Reviewed to 4eb740d.

README.md
prateek-kumar-improving: We already have all other documentation above. For now adding this here for consistency. 
Created an issue to update docs: https://github.com/valkey-io/valkey-glide-docs/issues/327

tests/TestValkeyGlide.php
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/commit/4d49bfcfccd0aefd8c1d8d2a244b09f1f1ddd680

valkey_glide.stub.php
prateek-kumar-improving: I kept `?int $node_discovery_mode = null` for two reasons:

1. **Consistency with the rest of `connect()`** — every parameter here follows the nullable `?type = null` convention (`?int $read_from = null`, `?bool $use_tls = null`, etc.). The C layer interprets `null` as "not provided → apply default," and `read_from` (the closest analog to this option) works exactly the same way. Making just this one parameter non-nullable would make it stand out from the other 22.

2. **Preserves the "unset" signal** — `null` lets the native layer distinguish "caller didn't pass it" from an explicit value, which is the pattern the whole constructor path relies on.

The default (STANDARD) is documented in the parameter's docblock so it's still discoverable from the API docs.

valkey_glide_core_commands.c
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-php/commit/f5e7756986c70372db458bdfffe500debc8b6915 — the mapping now has an explicit case for every enum value, and the final else throws a ValkeyGlideException

valkey_glide.c
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-php/commit/f5e7756986c70372db458bdfffe500debc8b6915

tests/ConnectionRequestTest.php
prateek-kumar-improving: Fixed in https://github.com/valkey-io/valkey-glide-php/commit/f5e7756986c70372db458bdfffe500debc8b6915

valkey_glide.c
prateek-kumar-improving: Fix in https://github.com/valkey-io/valkey-glide-php/commit/a101ece5f501f7e79fa3494a1cfb9090ec15b95e. node_discovery_mode is now consistently the last parameter across all five locations in valkey_glide.c

tests/ValkeyGlideNodeDiscoveryModeTest.php
prateek-kumar-improving: Fixed in commit: https://github.com/valkey-io/valkey-glide-php/commit/cccc50873bbf7fe28841ccd3a0bcdb321c3000f7

Added `assertConnected($client)` for the connectivity check; the GET verifies reads work in STATIC mode (ports the reference `connects_and_reads` test).

tests/ValkeyGlideNodeDiscoveryModeTest.php
prateek-kumar-improving: Kept as two tests to mirror the reference suites (Python/Node/Java split connects-and-reads from allows-writes).

tests/ValkeyGlideNodeDiscoveryModeTest.php
prateek-kumar-improving: Fixed in commit: https://github.com/valkey-io/valkey-glide-php/commit/cccc50873bbf7fe28841ccd3a0bcdb321c3000f7

Added `testStandardModeConnectsAndReads`, so each mode (STANDARD, STATIC, DISCOVER_ALL) now has explicit integration coverage.

tests/ValkeyGlideNodeDiscoveryModeTest.php
prateek-kumar-improving: Not required to trigger discovery — DISCOVER_ALL resolves the full topology at connect time, so the `CLIENT LIST` check would pass without any commands. The SET/GET are a deliberate secondary assertion (matching the Python, Node, and Java `discover_replicas_from_replica` tests): they confirm that a client bootstrapped from a *replica* address still routes writes to the discovered primary.

yipin-chen (approved):

PR #325, opened 2026-08-26
Update submodule glide core

### LINKED ISSUE #324: chore(core): bump valkey-glide submodule and update patch tooling (unblocks native path-based mTLS)
### Describe the feature

Bump the vendored `valkey-glide` submodule from the current pin (`1b0ffb14`, 2026-06-26) to a recent upstream `main`, and update the build tooling so it compiles cleanly. This unblocks native path-based mTLS (see "Motivation" below) and keeps the PHP client aligned with the current GLIDE core.

### Motivation

While implementing mTLS (#288 / PR #321), path-based client certificates (`client_cert_path` / `client_key_path`) were shipped as a **client-side one-time read** into the byte protobuf fields (`client_cert = 22`, `client_key = 23`), because the pinned core predates the native path fields.

Upstream glide-core (PR #6386, ~2026-07-21) added native fields:

- `client_cert_path = 31`
- `client_key_path = 32`
- `cert_reload = 33` (`ClientCertReloadConfig { bool enabled; optional uint32 interval_seconds; }`)

With these, the **core** reads the cert/key files at connection time and, when reload is enabled, periodically re-reads them so a rotated certificate is adopted on the next reconnect. This matches the Python/Node/Java clients' path-based mTLS semantics. The PHP client cannot offer true path semantics (core-side read + auto-reload) until the submodule is bumped to include fields 31/32/33.

### Blocker discovered

Simply bumping the submodule to latest `main` does **not** build. The regeneration script `utils/patch_proto_and_rust.py` strips proto3 `optional` from all fields (the vendored `protobuf-c` path requires this) and then hand-patches the generated Rust (`glide-core/src/client/types.rs`) to re-wrap the affected fields in `Some(...)` / `Option<...>`. That hand-patch list is hardcoded to the old field set.

The newer core adds `optional` fields the script does not know to re-wrap, producing compile errors in `glide-core`:

- `cert_reload.interval_seconds` (`optional uint32`) — used as `Option` (`.filter(...)`) in `types.rs`, but stripped to plain `u32`.
- `recovery_requests_queue_size` (proto field 34, `optional uint32`) — struct expects `Option<u32>`, gets `u32`.
- Resulting type-inference cascade errors on the new path-field handling in `types.rs`.

### Proposed Solution / Tasks

- [ ] Bump the `valkey-glide` submodule to a chosen recent upstream commit.
- [ ] Update `utils/patch_proto_and_rust.py` to add hand-patch rules for the newer `optional` fields (at minimum `cert_reload.interval_seconds` and `recovery_requests_queue_size`), and audit for any other newly-added `optional` fields.
- [ ] Regenerate `include/glide_bindings.h` (cbindgen) and `src/connection_request.pb-c.{c,h}` / `command_request.pb-c.{c,h}` / `response.pb-c.{c,h}`.
- [ ] Confirm `phpize && ./configure && make` builds cleanly with `-Werror`.
- [ ] Run the full `make test` suite (the ~2-month core jump affects behavior beyond TLS, so this needs the full CI matrix across supported PHP + engine versions).

### Follow-up (separate feature, depends on this)

Once the submodule includes fields 31/32/33, convert PHP's path-based mTLS to native semantics:

- Pass `client_cert_path` / `client_key_path` strings unread to proto fields 31/32 (stop reading files client-side).
- Populate `cert_reload` (field 33) with `enabled = true` and an optional `interval_seconds`, exposed via a new `cert_reload_interval_seconds` config key.
- Enforce mode exclusivity (path-based vs byte-based) and both-or-neither pairing per mode, mirroring the Python reference (`_validate_mtls_config` in `glide-shared/glide_shared/config.py`).
- Update stubs, README, and tests accordingly.

### Additional context

- Related feature: #288 (add mTLS), PR #321.
- Upstream core fields reference: `valkey-io/valkey-glide` `glide-core/src/protobuf/connection_request.proto` (fields 31/32/33, `ClientCertReloadConfig`).
- Root/CA certificate reload is out of scope upstream (tracked in valkey-io/valkey-glide#6529).

prateek-kumar-improving:
### Summary

Bumps the `valkey-glide` submodule from `1b0ffb14` to `d3feeaaf0` (upstream `main`) to keep the PHP client aligned with the current GLIDE core, and updates `utils/patch_proto_and_rust.py` so the newer core compiles.

### Issue link

Closes #324

- No user-facing API changes — core/tooling update only.

### Implementation

`patch_proto_and_rust.py` strips proto3 `optional` (required by the vendored `protobuf-c` path) and re-patches the generated Rust (`glide-core/src/client/types.rs`) to keep those fields `Option<...>`. The newer core added `optional` fields not in the script's hardcoded list, breaking the build. Added patch rules + verification for the newly added fields: `client_cert_path`, `client_key_path`, `cert_reload.interval_seconds`, and `recovery_requests_queue_size`.

Note: two commits are intentional (pointer bump; tooling fix). The submodule's in-tree edits are regenerated at build time and not committed — only the submodule commit pointer is tracked.

### Limitations

- Bumps the core and fixes the build only; no new client features are exposed in this PR.

### Testing

- Clean regeneration + Rust FFI build + PHP extension build (`-Werror`) succeed; `ConnectionRequest` tests pass (60/60). Full `make test` matrix should run in CI given the core jump.

### Checklist

- [ ] Tests are added or updated.

### DIFF
--- a/utils/patch_proto_and_rust.py
+++ b/utils/patch_proto_and_rust.py
@@ -173,6 +173,68 @@ def patch_rust_types_rs(rust_types_file):
             needs_patching = True
             log_message("Applied eviction_policy patch")
 
+        # Fix client_cert_path: changed from Option<Chars> to Chars (proto3 optional stripped).
+        # The generated field is now a plain `Chars`, so treat an empty string as "unset".
+        client_cert_path_pattern = (
+            r'let client_cert_path = value\s*'
+            r'\.client_cert_path\s*'
+            r'\.as_ref\(\)\s*'
+            r'\.filter\(\|p\| !p\.is_empty\(\)\)\s*'
+            r'\.map\(\|p\| p\.to_string\(\)\);'
+        )
+        client_cert_path_replacement = (
+            'let client_cert_path = Some(value.client_cert_path.to_string())'
+            '.filter(|p| !p.is_empty());'
+        )
+        if re.search(client_cert_path_pattern, new_content, re.DOTALL):
+            if not needs_patching:
+                create_backup(rust_types_file)
+            new_content = re.sub(
+                client_cert_path_pattern, client_cert_path_replacement, new_content, flags=re.DOTALL)
+            needs_patching = True
+            log_message("Applied client_cert_path patch")
+
+        # Fix client_key_path: changed from Option<Chars> to Chars (proto3 optional stripped).
+        client_key_path_pattern = (
+            r'let client_key_path = value\s*'
+            r'\.client_key_path\s*'
+            r'\.as_ref\(\)\s*'
+            r'\.filter\(\|p\| !p\.is_empty\(\)\)\s*'
+            r'\.map\(\|p\| p\.to_string\(\)\);'
+        )
+        client_key_path_replacement = (
+            'let client_key_path = Some(value.client_key_path.to_string())'
+            '.filter(|p| !p.is_empty());'
+        )
+        if re.search(client_key_path_pattern, new_content, re.DOTALL):
+            if not needs_patching:
+                create_backup(rust_types_file)
+            new_content = re.sub(
+                client_key_path_pattern, client_key_path_replacement, new_content, flags=re.DOTALL)
+            needs_patching = True
+            log_message("Applied client_key_path patch")
+
+        # Fix cert_reload.interval_seconds: changed from Option<u32> to u32.
+        cert_reload_interval_pattern = r'interval_seconds: proto_reload\.interval_seconds\.filter\(\|&s\| s != 0\),'
+        cert_reload_interval_replacement = 'interval_seconds: none_if_zero(proto_reload.interval_seconds),'
+        if re.search(cert_reload_interval_pattern, new_content):
+            if not needs_patching:
+                create_backup(rust_types_file)
+            new_content = re.sub(
+                cert_reload_interval_pattern, cert_reload_interval_replacement, new_content)
+            needs_patching = True
+            log_message("Applied cert_reload.interval_seconds patch")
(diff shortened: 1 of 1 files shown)


jamesx-improving (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+* Bump vendored `valkey-glide` submodule to align with the current GLIDE core, and update the protobuf/Rust patch tooling for newly added optional fields ([#325](https://github.com/valkey-io/valkey-glide-php/pull/325))
```
currantw: This should be an internal change only, so I'm not sure if we would normally include a `CHANGELOG` entry for this?

utils/patch_proto_and_rust.py
currantw, line 1: This approach to protobuf for the Rust client seems _very_ brittle and error prone. Not something we can solve in this pull request, but I certainly think this is something we should investigate and fix. It seems nearly inevitable to me that this will cause issues in the future. @prateek-kumar-improving can you raise an issue for this and initiate a discussion about this? I think this is something we should consider for the next release.

xShinnRyuu (approved):

CHANGELOG.md
prateek-kumar-improving: Updated. Removed from changelog.

utils/patch_proto_and_rust.py
prateek-kumar-improving: Created an issue for now. Will check before the next release if possible based on priority.
https://github.com/valkey-io/valkey-glide-php/issues/326

PR #335, opened 2026-09-16
feat(php): add READ_FROM_AZ_AFFINITY_ALL_NODES read strategy

### LINKED ISSUE #316: [Task] PHP: Add READ_FROM_AZ_AFFINITY_ALL_NODES read strategy
### Description

GLIDE added a new read strategy, `AZAffinityAllNodes`, in valkey-io/valkey-glide#6721, which closed valkey-io/valkey-glide#6653. Only the Python binding was wired up, so every other client needs the same surface. This issue tracks the PHP side.

The strategy spreads read requests equally, in round robin, across all nodes in the client's Availability Zone, primary and replicas alike. When no node in the client's AZ is available it falls back to a round robin across all nodes. It differs from `READ_FROM_AZ_AFFINITY_REPLICAS_AND_PRIMARY` in that it does not prioritise replicas ahead of the primary, which is what makes an even per-node read distribution possible on ElastiCache deployments that hit the 5 replica cap. The motivation is written up in valkey-io/valkey-glide#6653.

### Prerequisite

The GLIDE core this extension builds against has to include `AZAffinityAllNodes = 6` in `glide-core/src/protobuf/connection_request.proto`, which landed with valkey-io/valkey-glide#6721. Until the pinned core carries that value, the new constant has nothing to map to.

### Checklist

- [ ] Add `VALKEY_GLIDE_READ_FROM_AZ_AFFINITY_ALL_NODES` to `valkey_glide_read_from_t` in `common.h`. The enum currently stops at `3`, so the new member is `4`.
- [ ] Add the matching `READ_FROM_AZ_AFFINITY_ALL_NODES` class constant in `valkey_glide.stub.php`, with a docblock in the style of the neighbouring entries, then regenerate the arginfo header.
- [ ] Wire the new value through the connection request construction in `valkey_glide.c` and `valkey_glide_core_commands.c` so it reaches the core.
- [ ] Enforce the client AZ requirement for the new strategy, matching how `READ_FROM_AZ_AFFINITY` and `READ_FROM_AZ_AFFINITY_REPLICAS_AND_PRIMARY` are handled.
- [ ] Add coverage in `tests/ConnectionRequestTest.php`, which already enumerates the existing read-from values.
- [ ] Mention the new strategy in `examples/basic/configuration.php` alongside the existing ones.

### Additional Notes

The Python diff in valkey-io/valkey-glide#6721 is a close reference for the binding surface, including the exact wording of the validation errors and the docstrings.

Worth noting separately, and out of scope here: this extension's `valkey_glide_read_from_t` also has no member for the AZ agnostic `AllNodes` strategy (protobuf value 5) or for `LowestLatency`, both of which the core already supports. If you want those covered too they should get their own issue rather than riding along on this one.

The same parity gap in the other clients:

- Java valkey-io/valkey-glide#6790
- Node valkey-io/valkey-glide#6791
- Go valkey-io/valkey-glide#6792
- C# valkey-io/valkey-glide-csharp#533
- Ruby valkey-io/valkey-glide-ruby#294

prateek-kumar-improving:
### Summary

Adds the `READ_FROM_AZ_AFFINITY_ALL_NODES` read strategy to the PHP client, wiring up the binding surface for the `AZAffinityAllNodes` strategy (protobuf value `6`) that already exists in glide-core. The strategy spreads read requests equally, in round robin, across all nodes in the client's Availability Zone (primary and replicas alike), falling back to a round robin across all nodes when no in-AZ node is available. Unlike `READ_FROM_AZ_AFFINITY_REPLICAS_AND_PRIMARY`, it does not prioritize replicas ahead of the primary, which enables an even per-node read distribution on ElastiCache deployments that hit the 5 replica cap.

### Issue link

This Pull Request is linked to issue: [[Task] PHP: Add READ_FROM_AZ_AFFINITY_ALL_NODES read strategy](https://github.com/valkey-io/valkey-glide-php/issues/316)
Closes #316

- New `ValkeyGlide::READ_FROM_AZ_AFFINITY_ALL_NODES` (value `4`) read strategy, available for both standalone and cluster clients.
- AZ-affinity read strategies now require a client AZ. Selecting `AZ_AFFINITY`, `AZ_AFFINITY_REPLICAS_AND_PRIMARY`, or `AZ_AFFINITY_ALL_NODES` without a `client_az` throws a `ValkeyGlideException` at client creation instead of silently falling back to arbitrary nodes. A `client_az` that is empty or whitespace-only is treated as absent.

### Implementation

- `common.h`: added `VALKEY_GLIDE_READ_FROM_AZ_AFFINITY_ALL_NODES = 4` to `valkey_glide_read_from_t`.
- `valkey_glide.stub.php`: added the `READ_FROM_AZ_AFFINITY_ALL_NODES` class constant with a docblock describing the round-robin behaviour and fallback (arginfo header regenerated during build).
- `valkey_glide.c`: mapped input `read_from` value `4` to the new enum, and added AZ validation covering all three AZ-affinity strategies in the shared `valkey_glide_build_client_config_base` (so standalone, cluster, and monitor paths are all covered). A whitespace-only `client_az` resolves to absent so the strategy is rejected rather than engaged with a value that matches no node — only the borrowed PHP pointer is forwarded, never a copy, so config memory ownership is unchanged.
- `valkey_glide_core_commands.c`: mapped the new enum to `CONNECTION_REQUEST__READ_FROM__AZAffinityAllNodes` in `create_connection_request()`.
- `examples/basic/configuration.php`: documented the new value and the `client_az` requirement.
- `CHANGELOG.md`: added an Unreleased entry.

Reviewer note: the validation lives in the shared config-build helper rather than in each constructor, so both client types and the monitor path enforce it consistently.

### Limitations

- No read-only-mode incompatibility check (present in the Java/Go clients) — the PHP extension has no client-level read-only connection mode, so there is nothing to guard. The only "read-only" surface in PHP is read-only *commands* (`SORT_RO`, `EVAL_RO`, `FCALL_RO`).
- Validation rejects a blank/whitespace-only `client_az` but does not normalize (trim) a padded-but-valid value before forwarding it, since `client_az` is a borrowed PHP pointer and trimming would require changing its memory ownership. This is out of scope for the issue.
- Test coverage is at the connection-request/config level (via the mock, no server required), matching the issue checklist. No live AZ-routing integration test is included.

### Testing

- Added to `tests/ConnectionRequestTest.php`:
  - standalone + cluster protobuf mapping for `AZ_AFFINITY_ALL_NODES`,
  - standalone + cluster rejection when `client_az` is unset,
  - a parameterized check that all three AZ strategies require `client_az` (each error names its own strategy) on both client types,
  - a check that non-AZ strategies (`PRIMARY`, `PREFER_REPLICA`) do not require `client_az`,
  - a blank/whitespace-only `client_az` rejection check (`''`, `' '`, `'\t'`, `'\n'`, …).
- Full ConnectionRequestTest suite: 114 passed, 0 failed, 0 skipped.
- `make` builds the extension cleanly; `./lint-php.sh` passes.

### Checklist


valkey_glide.c
```diff
@@ -217,8 +217,25 @@ int valkey_glide_build_client_config_base(valkey_glide_php_common_constructor_pa
...
+     * config ownership is unchanged. */
+    config->client_az = NULL;
+    if (params->client_az && params->client_az_len > 0) {
+        for (size_t az_i = 0; az_i < params->client_az_len; az_i++) {
+            char az_ch = params->client_az[az_i];
+            if (az_ch != ' ' && az_ch != '\t' && az_ch != '\n' && az_ch != '\r' && az_ch != '\f' &&
+                az_ch != '\v') {
+                config->client_az = params->client_az;
+                break;
+            }
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/335/commits/2e42f8cbeac78fccef32728dd4fb5f0012df27ba

valkey_glide.stub.php
```diff
@@ -195,6 +195,14 @@ class ValkeyGlide
            */
     public const  READ_FROM_AZ_AFFINITY_REPLICAS_AND_PRIMARY = 3;
 
+          /**
+           *  @var int
+           * Spread the read requests equally among all nodes (primary and replicas) within the
+           * client's Availability Zone (AZ) in a round robin manner, falling back to a round robin
+           * across all nodes if no node in the client's AZ is available.
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/335/commits/4a084d0bda8dcfad095f3466ab7112fe0610248d

valkey_glide.c
```diff
@@ -217,8 +217,31 @@ int valkey_glide_build_client_config_base(valkey_glide_php_common_constructor_pa
...
+        for (size_t az_i = 0; az_i < params->client_az_len; az_i++) {
+            char az_ch = params->client_az[az_i];
+            /* Stop at the first NUL: bytes past it never reach the core. */
+            if (az_ch == '\0') {
+                break;
+            }
+            if (az_ch != ' ' && az_ch != '\t' && az_ch != '\n' && az_ch != '\r' && az_ch != '\f' &&
+                az_ch != '\v') {
+                config->client_az = params->client_az;
+                break;
```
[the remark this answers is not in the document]
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/335/commits/4a084d0bda8dcfad095f3466ab7112fe0610248d

valkey_glide.c
```diff
@@ -217,8 +217,39 @@ int valkey_glide_build_client_config_base(valkey_glide_php_common_constructor_pa
...
+                zend_throw_exception(get_valkey_glide_exception_ce(), az_error_message, 0);
+                return FAILURE;
+            }
+            if (az_ch != ' ' && az_ch != '\t' && az_ch != '\n' && az_ch != '\r' && az_ch != '\f' &&
+                az_ch != '\v') {
+                az_has_content = true;
+            }
+        }
+        if (az_has_content) {
+            config->client_az = params->client_az;
```
jamesx-improving, line 266: The blank half of this class is closed here — whitespace-only is rejected, and the NUL check is more than either the Node or Ruby sibling does. The padded half isn't: `"us-east-1a\n"`, e.g. from `file_get_contents` or an env var, has content, so it is forwarded raw here, matches no node under the core's exact-equality compare, and reads round-robin across all nodes — with no exception and no warning, because the core's `log_warn` fallback only fires when the AZ is absent.

**Suggested fix:** reject a value that differs from its trimmed form (no allocation, so the borrowed-pointer constraint you noted still holds), or trim into a config-owned copy as Node does.

Partially overlapping rather than new: a bot found the same class on the Node sibling (valkey-io/valkey-glide#7104, fixed there by normalizing) and it is open in Ruby as valkey-io/valkey-glide-ruby#318, which argues for rejecting instead of trimming. Nobody has raised it on this PR.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Changes
 
+* Add `READ_FROM_AZ_AFFINITY_ALL_NODES` read strategy for standalone and cluster clients ([#316](https://github.com/valkey-io/valkey-glide-php/issues/316))
```
jamesx-improving, line 11: This documents the addition but not the behaviour change: `AZ_AFFINITY` and `AZ_AFFINITY_REPLICAS_AND_PRIMARY` without a `client_az` now throw at client creation, where the core previously downgraded to `PreferReplica` with a warning. (greptile's summary asserts the changelog covers the `client_az` requirement — it doesn't.)

jamesx-improving (approved):
Both notes are inline and non-blocking. The AZ-normalization one is the substantive one and worth an explicit decision either way, since the sibling bindings disagree: Node's merged fix trims before forwarding, while valkey-glide-ruby#318 argues against rewriting the caller's value and would reject instead.

yipin-chen (approved):

CHANGELOG.md
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/335/commits/4b3ebf9eb734e896b281a36a311b04404f12291f

valkey_glide.c
prateek-kumar-improving: Fixed in this commit: https://github.com/valkey-io/valkey-glide-php/pull/335/commits/4b3ebf9eb734e896b281a36a311b04404f12291f

